PR #3, opened 2019-08-15
InjBO optimization

murilobalves:


apsuite/optimization.py
```diff
@@ -63,30 +61,11 @@ def ndim(self):
     def ndim(self, value):
```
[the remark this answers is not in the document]
murilobalves: Done!

apsuite/optimization.py
```diff
@@ -75,92 +131,159 @@ def _create_swarm(self):
...
-    def _check_initialization(self):
-        if len(self._upper_limits) != len(self._lower_limits):
-            print('Warning: Upper and Lower Limits has different lengths')
+    def get_change(self):
+        pass
 
-        if self._ndim != len(self._upper_limits):
-            print('Warning: Dimension incompatible with limits!')
+    def set_change(self):
+        pass
```
xresende: I usually implement an abstrat base method like this:

```python
def set_change(self):
    """DocString."""
    raise NotImplementedError
```
I think it is safer to raise an exception when the abstract method is invoked. Otherwise the class user may be using the base class improperly without knowing it (and eventual bugs might be lurking in...)

xresende: I suppose `set_change` is an abstract method to be implemented in a derived class... is this so?

murilobalves: Yes, correct. Thanks!

PR #41, opened 2020-03-20
Tune correction

murilobalves:


apsuite/commissioning_scripts/calc_tune_matrix.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python-sirius
```
xresende: You should drop she-bang line in python import modules.

apsuite/commissioning_scripts/calc_tune_matrix.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def get_kl(self, model=None, quads=None):
+        """."""
+        if model is None:
+            model = self.model
+        if quads is None:
+            quads = self.quads
+        kl = []
+        for q in quads:
+            kl.append(model[self.fam[q]['index'][0][0]].KL)
```
fernandohds564: I think this could be more generic, considering families with an arbitrary number of elements and elements with an arbitrary number of segments.

apsuite/commissioning_scripts/calc_tune_matrix.py
```diff
@@ -0,0 +1,108 @@
...
+        inv_s[np.isinf(inv_s)] = 0
+        inv_s = np.diag(inv_s)
+        inv_matrix = np.dot(np.dot(v.T, inv_s), u.T)
+        tunex0, tuney0 = self.get_tunes(model)
+        print(tunex0, tuney0)
+        tunex_new, tuney_new = tunex0, tuney0
+        kl = self.get_kl(model)
+        tol = 1e-6
+
+        while abs(tunex_new - tunex) > tol or abs(tuney_new - tuney) > tol:
```
fernandohds564: I think it is better to introduce a maximum number of iterations for this loop. Although it is not likely, this could turn into an infinite loop.

apsuite/commissioning_scripts/calc_tune_matrix.py
```diff
@@ -0,0 +1,108 @@
...
+        """."""
+        if model is None:
+            model = self.model
+        if quads is None:
+            quads = self.quads
+        if kl is None:
+            raise Exception('Missing KL values')
+        newmod = _dcopy(model)
+        for idx, q in enumerate(quads):
+            newmod[self.fam[q]['index'][0][0]].KL = kl[idx]
```
fernandohds564: same here

apsuite/commissioning_scripts/calc_tune_matrix.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python-sirius
+"""."""
+
+from copy import deepcopy as _dcopy
+import numpy as np
+from pymodels import bo, si
+import pyaccel
+
+
+class Tune():
```
fernandohds564: I think it is important to generalize this, allowing the user to define which quadrupole families will be used for correction. For example, this class should be generic enough to allow the user to correct the tunes with dipole families, or sextupole families... We cannot predict the case of use of such a general purpose class.

In this line of thought, this class miss a property with the name of the quadrupole families which will be used and also, maybe optional initialization attributes with the desired quadrupole families.

Besides, it is important to reproduce in this class the proportional algorithm which is used in our tune correction IOC. This method groups QDs and QFs in groups and apply DeltaKs proportional to the strength of the given family. To accomplish this in a generic way it would be nice to have properties to set the QDs and QFs in two groups...

apsuite/commissioning_scripts/calc_tune_matrix.py
murilobalves: ok, done.

apsuite/commissioning_scripts/calc_tune_matrix.py
murilobalves: done.

apsuite/commissioning_scripts/calc_tune_matrix.py
murilobalves: done.

apsuite/commissioning_scripts/calc_tune_matrix.py
murilobalves: done.

apsuite/commissioning_scripts/calc_tune_matrix.py
murilobalves: Indeed. I believe that the changes I've made so far are converging to the idea of using generic elements as knobs for tune correction (for example, setting in the initialization the input knobs_names as a list with magnets names to be used).

The proportional algorithm implementation is on top of my to-do list.

murilobalves: done!

apsuite/optics_analysis/tune_correction.py
```diff
@@ -0,0 +1,200 @@
...
+        """."""
+        self.model = model
+        self.acc = acc
+        if acc == 'BO':
+            if knobs_names is None:
+                self.knobs_names = self.BOKNOBS
+            self.fam = bo.get_family_data(model)
+        elif acc == 'SI':
+            if knobs_names is None:
+                self.knobs_names = self.SIKNOBS
```
anacso17: I think it is missing to handle the case the user create `TuneCorr` class passing `knobs_names` parameter

murilobalves: Sure! It is corrected now, thanks.

apsuite/optics_analysis/tune_correction.py
```diff
@@ -0,0 +1,244 @@
...
+    @property
+    def ALL(self):
+        """."""
+        return self.QFs + self.QDs
+
+
+class TuneCorr():
+    """."""
+
+    SIDEFKNOBS = KnobTypes(
```
xresende: I think that using underscore separators, like `SI_DEF_KNOBS`, makes code much more readable. And with code completion you don't even have to type awkward `_` char... don't you think?

fernandohds564: ok @xresende. I'll change it.

fernandohds564: done!

PR #44, opened 2020-03-25
Base optics correction class

murilobalves:
- Includes new base optics correction class
- Refactor tune correction code to use this new class
- Includes new chromaticity correction class


apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __init__(self, model=None, acc=None, method=None, grouping=None):
+        """."""
+        self.model = model
+        self.acc = acc
+        self._method = BaseCorr.METHODS.Proportional
+        self._grouping = BaseCorr.GROUPING.TwoKnobs
+        self.knobs = None
+        self.fam = None
+        self.strength_type = None
```
fernandohds564: I think it would be better if `strength_type` were class attribute, not an object attribute!

apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
...
+    def method_str(self):
+        """."""
+        return BaseCorr.METHODS._fields[self._method]
+
+    def define_knobs(self, f_knobs, d_knobs, strength_type):
+        """."""
+        self.knobs = KnobTypes(Focusing=f_knobs, Defocusing=d_knobs)
+        self.strength_type = strength_type
+
+    def get_parameter(self, model=None):
```
fernandohds564: maybe this could be a hidden method. The classes which will derive from this one could implement it and also define another method to get the appropriate parameter.  For example the TuneCorr class would define something like this:

```
def _get_parameter(self, model=None):
    implement here  the real tune calculation

def get_tunes(self, model=None):
    return self._get_parameter(model=model)
```

this way each derived class would have a method with the appropriate name.

apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
...
+        """."""
+        raise NotImplementedError
+
+    def calc_jacobian_matrix(self, model=None):
+        """."""
+        raise NotImplementedError
+
+    def get_strength(self, model=None, knobs=None):
```
fernandohds564: same here.

apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
...
+            self._method = int(value in BaseCorr.METHODS._fields[1])
+        elif int(value) in BaseCorr.METHODS:
+            self._method = int(value)
+
+    @property
+    def method_str(self):
+        """."""
+        return BaseCorr.METHODS._fields[self._method]
+
+    def define_knobs(self, f_knobs, d_knobs, strength_type):
```
fernandohds564: With the suggestion made above, regarding the attribute `strength_type`, I think this method is not necessary. The knobs can be defined in the `__init__` method of the derived class.

apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
+"""."""
+
+from collections import namedtuple as _namedtuple
+from copy import deepcopy as _dcopy
+import numpy as np
+
+__KNOBTYPES = _namedtuple('KnobTypes', ['Focusing', 'Defocusing'])
+
+
+class KnobTypes(__KNOBTYPES):
```
xresende: Pylint complains about namedtuple inheretence (`inherent-non-class`)
In my opinion we should avoid it and should use simpler mechanisms.

https://stackoverflow.com/questions/42385916/inheriting-from-a-namedtuple-base-class-python
http://zecong.hu/2019/08/10/inheritance-for-namedtuples/

apsuite/optics_analysis/base_correction.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            jmat = _dcopy(jacobian_matrix)
+
+        nominal_stren = self.get_strength(model)
+        if self._method == BaseCorr.METHODS.Proportional:
+            jmat *= nominal_stren
+        if self._grouping == BaseCorr.GROUPING.TwoKnobs:
+            jmat = self._group_2knobs_matrix(jmat)
+
+        U, S, V = np.linalg.svd(jmat, full_matrices=False)
```
xresende: Pylint requires variable names with three or more letters.
In other python modules I used `umat`, `smat`, `vmat`... I did not like these names but this is what I come up compatible with pylint. Do you have suggestions? Whatever names we choose I vote for using the same names across python modules for variables which represent the same idea. I think this *straitjacket* makes code maintenance much easier for original and new-comer developers.

apsuite/optics_analysis/base_correction.py
xresende: @fernandohds564

apsuite/optics_analysis/base_correction.py
murilobalves: good suggestion. done!

apsuite/optics_analysis/base_correction.py
murilobalves: ok!

apsuite/optics_analysis/base_correction.py
murilobalves: ok.

apsuite/optics_analysis/base_correction.py
murilobalves: I think `umat`, `smat`, `vmat` is ok. I changed here and I will standardize in other modules which uses SVD.

apsuite/optics_analysis/base_correction.py
fernandohds564: ok I agree! I was thinking about changing this. We could create a simple class with object attributes `focusing` and `defocusing` and a property named `all` which combines both.

fernandohds564: done!

PR #45, opened 2020-03-27
Coupling correction

murilobalves:


apsuite/optics_analysis/coupling_correction.py
```diff
@@ -105,56 +108,58 @@ def set_ksl(self, model=None, skewidx=None, ksl=None):
         return newmod
 
     @staticmethod
-    def get_fm(res):
+    def get_figm(res):
         """Calculate figure of merit from residue vector."""
         return np.sqrt(np.sum(np.abs(res)**2)/res.size)
```
fernandohds564: Why did you have to take the `np.abs` of `res` here?
I think that for long arrays `res*res` is faster than `res**2`.

murilobalves: my bad!

apsuite/optics_analysis/coupling_correction.py
```diff
@@ -82,79 +85,78 @@ def get_ksl(self, model=None, skewidx=None):
...
-            raise Exception('Missing KsL values')
-        newmod = _dcopy(model)
+        if delta_ksl is None:
+            raise Exception('Missing Delta KsL values')
         for idx_mag, mag in enumerate(skewidx):
-            for idx_seg, seg in enumerate(mag):
-                newmod[seg].KsL = ksl[idx_mag][idx_seg]
-        return newmod
+            delta = delta_ksl[idx_mag]/len(mag)
+            for _, seg in enumerate(mag):
```
fernandohds564: you don't need `enumerate` here.

apsuite/optics_analysis/coupling_correction.py
```diff
@@ -82,79 +85,78 @@ def get_ksl(self, model=None, skewidx=None):
...
+            return CouplingCorr.CORR_STATUS.Fail
+        return CouplingCorr.CORR_STATUS.Sucess
 
     def coupling_correction(self,
                             model,
-                            matrix=None, nsv=None, niter=10, tol=1e-6,
+                            jacobian_matrix=None,
+                            nsv=None, nr_max=10, tol=1e-6,
                             res0=None,
                             method='orbrespm'):
```
fernandohds564: I think `method` should be an object property, not an input argument.

apsuite/optics_analysis/coupling_correction.py
murilobalves: indeed.

apsuite/optics_analysis/coupling_correction.py
murilobalves: ok.

apsuite/optics_analysis/coupling_correction.py
```diff
@@ -13,11 +13,13 @@ class CouplingCorr():
...
     CORR_STATUS = _namedtuple('CorrStatus', ['Fail', 'Sucess'])(0, 1)
 
-    def __init__(self, model, acc, dim='4d', skew_list=None):
+    def __init__(self, model, acc, dim='4d',
+                 skew_list=None, method='orbrespm'):
         """."""
         self.model = model
         self.acc = acc
         self.dim = dim
+        self.method = method
```
fernandohds564: I think it is nice to create a namedtuple for the possible methods:

`METHODS = _namedtuple('Methods', ['Additional', 'Proportional'])(0, 1)`

keep the method attribute as private object attribute (`self._method`) and create a property for it, just like it is done for the optics correction.

apsuite/optics_analysis/coupling_correction.py
fernandohds564: see comment below

apsuite/optics_analysis/coupling_correction.py
murilobalves: done!

PR #46, opened 2020-03-27
Optics correction

murilobalves:
- LOCO-like optics correction
- All quadrupoles used individually as default
- Additional and proportional methods
- The code minimizes the difference of residue vector which includes betas at BPMs e correctors (betax@CH and betay@CV), horizontal dispersion at BPMs and tunes.
- It is intended that in the next version this code will include optics symmetrization based on the code implemented in AT.

### DIFF
--- a/apsuite/optics_analysis/optics_correction.py
+++ b/apsuite/optics_analysis/optics_correction.py
@@ -146,7 +146,7 @@ class OpticsCorr():
         return np.sqrt(np.sum(res*res)/res.size)
 
     def optics_corr_loco(self,
-                         model,
+                         model=None,
                          goal_model=None,
                          jacobian_matrix=None,
                          nsv=None, nr_max=10, tol=1e-6):
@@ -156,6 +156,8 @@ class OpticsCorr():
         and minimizes the residue vector:
         Delta [Bx@BPM, Bx@CH, By@BPM, By@CV, Dx@BPM, Tunex, Tuney].
         """
+        if model is None:
+            model = self.model
         if jacobian_matrix is None:
             jmat = self.calc_jacobian_matrix(model)
         else:
@@ -195,7 +197,7 @@ class OpticsCorr():
         return OpticsCorr.CORR_STATUS.Sucess
 
     def optics_correction(self,
-                          model,
+                          model=None,
                           goal_model=None,
                           jacobian_matrix=None,
                           nsv=None, nr_max=10, tol=1e-6,
@@ -205,6 +207,8 @@ class OpticsCorr():
         Methods available:
         - LOCO-like correction
         """
+        if model is None:
+            model = self.model
         if method == 'loco':
             result = self.optics_corr_loco(
                 model=model, goal_model=goal_model,
--- a/apsuite/optics_analysis/optics_correction.py
+++ b/apsuite/optics_analysis/optics_correction.py
@@ -141,11 +141,11 @@ class OpticsCorr():
         """Calculate figure of merit from residue vector."""
         return np.sqrt(np.sum(np.abs(res)**2)/res.size)
 
-    def optics_corr(self,
-                    model,
-                    goal_model=None,
-                    jacobian_matrix=None,
-                    nsv=None, nr_max=10, tol=1e-6):
+    def optics_corr_loco(self,
+                         model,
+                         goal_model=None,
+                         jacobian_matrix=None,
+                         nsv=None, nr_max=10, tol=1e-6):
         """Optics correction LOCO-like.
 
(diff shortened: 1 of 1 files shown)


apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,217 @@
...
+        if model is None:
+            model = self.model
+        if knobsidx is None:
+            knobsidx = self.knobs_idx
+        kl = []
+        for mag in knobsidx:
+            kl_seg = []
+            for seg in mag:
+                kl_seg.append(model[seg].KL)
+            kl.append(kl_seg)
```
fernandohds564: I think you have to sum the KL of the segments here, don't you?

apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,217 @@
...
+        return np.array(kl)
```
fernandohds564: If you don't sum the KL as commented above. Here you are assuming all knobs have the same number of segments. I know that for the current version of Sirius lattice this is true, but should we make this assumption here?

apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,217 @@
...
+
+        res0 = self._get_optics_vector(model)
+        self.jacobian_matrix = np.zeros((res0.size, len(self.knobs_idx)))
+        delta = 1e-6
+
+        modcopy = model[:]
+        for idx, nmag in enumerate(self.knobs_idx):
+            dlt = delta/len(nmag)
+            for seg in nmag:
+                modcopy[seg].KL += dlt
```
fernandohds564: with this implementation, the total `KL` you are applying to each element is `n*dlt` where `n` is the number of segments of the element. Is really that what you want?

apsuite/optics_analysis/optics_correction.py
murilobalves: yes! done.

apsuite/optics_analysis/optics_correction.py
murilobalves: fixed!

apsuite/optics_analysis/optics_correction.py
murilobalves: yes! 'dlt = delta/len(nmag)' is the division of the total delta in 'N' parts where 'N' is the number of segments. Then I apply 'dlt' for each one of the 'N' segments, therefore obtaining the total 'delta' for each element as desired.

apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            return OpticsCorr.CORR_STATUS.Fail
+        return OpticsCorr.CORR_STATUS.Sucess
+
+    def optics_correction(self,
+                          model,
+                          goal_model=None,
+                          jacobian_matrix=None,
+                          nsv=None, nr_max=10, tol=1e-6,
+                          method='loco'):
```
fernandohds564: I think this `method`option should be an object property, not an input argument.

murilobalves: done.

apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,220 @@
...
+    METHODS = _namedtuple('Methods', ['Additional', 'Proportional'])(0, 1)
+    CORR_STATUS = _namedtuple('CorrStatus', ['Fail', 'Sucess'])(0, 1)
+
+    def __init__(self, model, acc, dim='4d', knobs_list=None,
+                 method=None, correction_method='loco'):
+        """."""
+        self.model = model
+        self.acc = acc
+        self.dim = dim
+        self.corr_method = correction_method
```
fernandohds564: treat it as a hidden attribute with a property, just like the `method` attribute.

apsuite/optics_analysis/optics_correction.py
```diff
@@ -0,0 +1,220 @@
...
+                          jacobian_matrix=None,
+                          nsv=None, nr_max=10, tol=1e-6):
+        """Optics correction method selection.
+
+        Methods available:
+        - LOCO-like correction
+        """
+        if model is None:
+            model = self.model
+        if self.corr_method == 'loco':
```
fernandohds564: here, the comparison should be against the `namedtuple` you will define.

apsuite/optics_analysis/optics_correction.py
fernandohds564: see comments below.

PR #47, opened 2020-03-27
Orbit correction

murilobalves:


apsuite/optics_analysis/orbit_correction.py
```diff
@@ -0,0 +1,164 @@
...
+            corridx = self.corr_idx
+        if kicks is None:
+            raise Exception('Missing Kicks values')
+        newmod = _dcopy(model)
+        for idx_mag, mag in enumerate(corridx):
+            for idx_seg, seg in enumerate(mag):
+                kick = kicks[idx_mag][idx_seg]
+                if idx_mag < self._nch:
+                    if np.abs(kick) > OrbitCorr.MAX_HKICK:
+                        kick = np.sign(kick) * OrbitCorr.MAX_HKICK
```
fernandohds564: I think this won't work well. When you reach the MAX_KICK you have three options: 
 - scale all kicks in such a way that the maximum value is equal to MAX_KICK;
 - give up, raise an error or whatever, so that the user knows about this and remove by hand that corrector from the correction;
 - remove the corrector from the matrix, and calculate the inverse once again.

I think the easiest way is to give up and let the user handle this problem.

apsuite/optics_analysis/orbit_correction.py
```diff
@@ -0,0 +1,164 @@
...
+                    newmod[seg].vkick_polynom = kick
+        return newmod
+
+    @staticmethod
+    def get_figm(res):
+        """Calculate figure of merit from residue vector."""
+        return np.sqrt(np.sum(res*res)/res.size)
+
+    def orbit_corr(self,
+                   model,
```
fernandohds564: I think model should be optional here. Just like in tune or chrom corrections, the default should be `self.model`

apsuite/optics_analysis/orbit_correction.py
```diff
@@ -0,0 +1,178 @@
...
+        kicksx_above = kicksx > OrbitCorr.MAX_HKICK
+        kicksy_above = kicksy > OrbitCorr.MAX_VKICK
+
+        if sum(kicksx_above) or sum(kicksy_above):
+            str_above = '\n'
+            str_above += 'HKick > MaxHKick at CHs: {0:s} \n'
+            str_above += 'VKick > MaxVKick at CVs: {1:s}'
+            xlist = str(np.argwhere(kicksx_above).flatten())
+            ylist = str(np.argwhere(kicksy_above).flatten())
+            raise Exception(str_above.format(xlist, ylist))
```
fernandohds564: I think it is better to raise a `ValueError`. `Exception` is too general.

apsuite/optics_analysis/orbit_correction.py
```diff
@@ -0,0 +1,178 @@
...
+        orb = self._get_orbit_residue(model)
+        dorb = orb - goal_orbit
+        bestfigm = OrbitCorr.get_figm(dorb)
+        if bestfigm < tol:
+            return OrbitCorr.CORR_STATUS.Sucess
+
+        for _ in range(nr_max):
+            dkicks = np.dot(ismat, dorb)
+            self._set_delta_kicks(model=model, delta_kicks=dkicks)
+            self._check_kicks(model=model)
```
fernandohds564: you are applying the kicks and then checking them against the maximum values, but you don't remove the delta applied before raising the error. Isn't it better to remove or don't even apply the exceeding kicks?

apsuite/optics_analysis/orbit_correction.py
murilobalves: Ok, now this case is handled by the new `_check_kicks` method.

apsuite/optics_analysis/orbit_correction.py
murilobalves: when I was implementing I thought about this question and here is what I decided:

I had the "let the user handle it" policy in my mind. If the kicks are greater than the maximum the user must restart the correction with a different singular values or corrector selection, beginning with the initial kicks again. 

The chance that before the kicks surpasses the max value they are already close to the max is high, then keeping the almost spoiled model may not be very useful. 

In the case that the variations are big and suddenly the kicks overcomes the max, it is an indicative that the singular values should be reduced and the correction must restart anyway.

what do you think?

apsuite/optics_analysis/orbit_correction.py
fernandohds564: see comment below.

apsuite/optics_analysis/orbit_correction.py
murilobalves: Applying the kicks and then checking them makes easier (at least to me) to get the list of corrector whose kicks are greater than max.

fernandohds564: ok! lets do this way then!

apsuite/optics_analysis/orbit_correction.py
murilobalves: ok, changed.

PR #49, opened 2020-04-07
Fix apsuite

murilobalves:
- Fix codes to follow changes made in lnls-fac/pyaccel#59
- Follow pylint recommendations in some codes

### DIFF
--- a/apsuite/commissioning_scripts/calc_orbcorr_mat.py
+++ b/apsuite/commissioning_scripts/calc_orbcorr_mat.py
@@ -20,8 +20,8 @@ class OrbRespmat():
             raise Exception('Set models: BO or SI')
         self.dim = dim
         self.bpms = self._get_idx(self.fam_data['BPM']['index'])
-        self.ch = self._get_idx(self.fam_data['CH']['index'])
-        self.cv = self._get_idx(self.fam_data['CV']['index'])
+        self.chidx = self._get_idx(self.fam_data['CH']['index'])
+        self.cvidx = self._get_idx(self.fam_data['CV']['index'])
 
     @staticmethod
     def _get_idx(indcs):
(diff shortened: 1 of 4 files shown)


fernandohds564:
thanks for the PR @murilobalves !

apsuite/commissioning_scripts/calc_orbcorr_mat.py
```diff
@@ -20,8 +20,8 @@ def __init__(self, model, acc, dim='4d'):
             raise Exception('Set models: BO or SI')
         self.dim = dim
         self.bpms = self._get_idx(self.fam_data['BPM']['index'])
-        self.ch = self._get_idx(self.fam_data['CH']['index'])
-        self.cv = self._get_idx(self.fam_data['CV']['index'])
+        self.chidx = self._get_idx(self.fam_data['CH']['index'])
```
xresende, line 23: I think these types (`chidx`, `cvidx`) will be frequently recurring objects across our libraries. I think we should carefully choose names and use them consistently, if possible. Our libraries will be more intuitive this way. This applies to many categories (`ch`, `psname`, `bpm`, `element`, etc). I propose `ch_idx`, for example, as word separation makes names clearer and it is being used in other repos.

also, `bpm` in this code should follow the same name logic as `ch` and `cv`. I usually reserve names like `bpm`, `pwrsupply`, `pvs` to objects of the corresponding types, not associated indices. I think standards like this make code easier to read, and interfaces more intuitive to use. what do you think?

apsuite/commissioning_scripts/calc_orbcorr_mat.py
```diff
@@ -32,54 +32,56 @@ def _calc_rfline(self):
             self.model, 'pass_method', 'cavity_pass')[0]
         rffreq = self.model[idx].frequency
         if self.dim == '6d':
-            df = 10
-            self.model[idx].frequency = rffreq + df
+            dfreq = 10
```
xresende, line 35: I propose moving this hard-coded constant to a class constant attribute, at the top of the class implementation, where it is much more visible.

```python 3
class OrbRespMat:
    """."""

    _FREQ_RESP = 10.0  # [Hz]
```

apsuite/commissioning_scripts/calc_orbcorr_mat.py
```diff
@@ -32,54 +32,56 @@ def _calc_rfline(self):
...
-            self.model[idx].frequency = rffreq - df
+            self.model[idx].frequency = rffreq - dfreq
             orbn = pyaccel.tracking.find_orbit6(self.model, indices='open')
             self.model[idx].frequency = rffreq
-            rfline = (orbp[[0, 2], :] - orbn[[0, 2], :])/2/df
+            rfline = (orbp[[0, 2], :] - orbn[[0, 2], :])/2/dfreq
             rfline = rfline[:, self.bpms].flatten()
         else:
-            de = 1e-5
+            denergy = 1e-5
```
xresende, line 44: same comment here as before.

PR #51, opened 2020-04-09
LOCO Upgrade

murilobalves:
- LOCOConfig: Now the algorithm setup is selected with the help of class constants in namedtuple format:
    - INVERSION: Normal or Transpose - Select if the jacobian to be inverted will be J or J.T * J
    - MINIMIZATION: Gauss-Newton or Levenberg-Marquardt - Select the minimization algorithm to be used, note: Levenberg-Marquardt works only using transpose inversion
    - SVD: Selection or Threshold - Select if the singular values selection in the jacobian will be by selecting the number of SV to keep or removing the low SV which are lower than SV_max * Threshold

- LOCO: The SV selection approach is changed: the old code selected the SV of 1/S matrix, where S is singular matrix found by J = U * S * V.T in the normal inversion case and in the transpose inversion case S is obtained by J.T * J = U * S * V.T. 
Now this code selects the SV of the S itself, then reconstructs the jacobian with this filtered singular matrix. After that, no other SV selection must be done. This is useful to keep the normal and transpose inversion methods equivalent with respect of singular values selection, which seems to be the proper approach.

### DIFF
--- a/apsuite/loco/config.py
+++ b/apsuite/loco/config.py
@@ -1,5 +1,6 @@
 """."""
 
+from collections import namedtuple as _namedtuple
 from copy import deepcopy as _dcopy
 import numpy as np
 import pyaccel
@@ -10,9 +11,6 @@ from apsuite.loco.utils import LOCOUtils
 class LOCOConfig:
     """LOCO configuration template."""
 
-    SVD_METHOD_SELECTION = 0
-    SVD_METHOD_THRESHOLD = 1
-
     DEFAULT_DELTA_K = 1e-6  # [1/m^2]
     DEFAULT_DELTA_KS = 1e-6  # [1/m^2]
     DEFAULT_DELTA_DIP_KICK = 1e-6  # [rad]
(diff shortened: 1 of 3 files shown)


apsuite/loco/main.py
```diff
@@ -16,8 +16,8 @@ class LOCO:
     DEFAULT_TOL = 1e-3
```
xresende, line 16: these constants have units? if they so, I think we should include a comment in the line with it. I always use the format
```python

FREQ_DELTA = 10.0  # [Hz]

File: apsuite/loco/config.py [hunk 2]
```diff
@@ -21,8 +19,18 @@ class LOCOConfig:
...
     FAMNAME_RF = 'SRFCav'
 
+    INVERSION = _namedtuple(
+        'Methods', ['Normal', 'Transpose'])(0, 1)
+    MINIMIZATION = _namedtuple(
+        'Methods', ['GaussNewton', 'LevenbergMarquardt'])(0, 1)
+    SVD = _namedtuple(
+        'Methods', ['Selection', 'Threshold'])(0, 1)
+
     def __init__(self, **kwargs):
```
xresende, line 29: nice. I think the interface to LOCO config is becoming very clean.

apsuite/loco/main.py
xresende: @murilobalves , I don't know why but when I make a comment on a code line, the line that shows ups in this conversation is not necessarily the correct one. I must be misusing it.

I was looking at `DEFAULT_DELTAK_NORMALIZATION`... but please check all constants.

murilobalves: Actually `DEFAULT_DELTAK_NORMALIZATION` is a parameter I would like to discuss with the group. I guess its unit will also be a result from this discussion. I just set it to 1.0 temporarily. The other class constants in LOCO are dimensionless. The constants in LOCOConfig has a comment indicating its units as you suggested.

xresende: all right, let's discuss in gitter... aproving this PR.

PR #52, opened 2020-04-13
Add FitInjTraj class

fernandohds564:
Usage:
```
import numpy as np
from apsuite.commissioning_scripts.inj_traj_fitting import FitInjTraj

np.random.seed(42)

fit_traj = FitInjTraj()

x0, xl0, y0, yl0, de0 = -9.0e-3, 0.0e-3, 0.0e-3, 0.0, 0.01
trajx, trajy, trajsum = fit_traj.simulate_sofb(x0, xl0, y0, yl0, de0)

# trajx, trajy, trajsum = fit_traj.get_traj_from_sofb()

vecs = fit_traj.do_fitting(trajx, trajy, tol=1e-8)
```

### DIFF
--- a/apsuite/commissioning_scripts/inj_traj_fitting.py
+++ b/apsuite/commissioning_scripts/inj_traj_fitting.py
@@ -149,6 +149,8 @@ class FitInjTraj(BaseClass):
         maxidx = np.sum(indcs)
         trajx = trajx[:maxidx]
         trajy = trajx[:maxidx]
+        trajx *= 1e-6  # from um to m
+        trajy *= 1e-6  # from um to m
         return trajx, trajy, summ
 
     def simulate_sofb(
--- a/apsuite/commissioning_scripts/inj_traj_fitting.py
+++ b/apsuite/commissioning_scripts/inj_traj_fitting.py
@@ -98,12 +98,15 @@ class FitInjTraj(BaseClass):
         de_ini = np.mean(trajx) / etax_ave
         return np.array([x_ini, xl_ini, y_ini, yl_ini, de_ini])
 
-    def do_fitting(self, trajx, trajy, vec0=None, max_iter=5, tol=1e-6):
+    def do_fitting(
+            self, trajx, trajy, vec0=None, max_iter=5, tol=1e-6, full=False):
         """."""
         vec0 = vec0 if vec0 is not None else self.calc_init_vals(trajx, trajy)
         res0 = self.calc_residue(vec0, trajx, trajy)
         chi0 = self.calc_chisqr(res0)
         vecs = [vec0, ]
+        residues = [res0, ]
+        chis = [chi0, ]
 
         vec, res = vec0.copy(), res0.copy()
         factor = 1
@@ -123,12 +126,17 @@ class FitInjTraj(BaseClass):
             else:
                 vec0 = vec.copy()
                 res0 = res.copy()
-                vecs.append(vec0)
                 chi0 = chi
+                vecs.append(vec0)
+                residues.append(res0)
+                chis.append(chi0)
                 factor = min(factor*2, 1)
             if chi0 < tol:
                 break
-        return vecs
+        if full:
+            return vecs, residues, chis
+        else:
+            return vecs
 
     def get_traj_from_sofb(self):
         """."""
--- a/apsuite/commissioning_scripts/inj_traj_fitting.py
+++ b/apsuite/commissioning_scripts/inj_traj_fitting.py
@@ -21,6 +21,7 @@ class Params:
         self.simul_npart = 1000
         self.simul_cutoff = 6
         self.count_rel_thres = 1/3
+        self.count_init_ref = 3
 
 
 class FitInjTraj(BaseClass):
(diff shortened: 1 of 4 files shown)


apsuite/commissioning_scripts/inj_traj_fitting.py
```diff
@@ -0,0 +1,230 @@
...
+                break
+        return vecs
+
+    def get_traj_from_sofb(self):
+        """."""
+        trajx = self.devices['sofb'].trajx.copy()
+        trajy = self.devices['sofb'].trajy.copy()
+        summ = self.devices['sofb'].sum.copy()
+
+        ini = np.mean(summ[:3])
```
murilobalves, line 125: maybe this could be an input with default value of 3 or an attribute of `Params` just like `count_rel_thres`, since these two parameters seems to be closely related.

apsuite/commissioning_scripts/inj_traj_fitting.py
```diff
@@ -0,0 +1,230 @@
...
+    def calc_jacobian(self, vec, size=160):
+        """."""
+        mat = np.zeros((2*size, 5))
+        zer = np.zeros(size)
+        dlt = 1e-5
+        for i in range(vec.size):
+            dvec = np.array(vec)
+            dvec[i] += dlt/2
+            res_pos = self.calc_residue(dvec, zer, zer)
+            dvec[i] -= dlt
```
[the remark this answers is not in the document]
murilobalves: ops, my bad: +1/2 - 1 = -1/2, ok!

apsuite/commissioning_scripts/inj_traj_fitting.py
```diff
@@ -0,0 +1,230 @@
...
+                vec0 = vec.copy()
+                res0 = res.copy()
+                vecs.append(vec0)
+                chi0 = chi
+                factor = min(factor*2, 1)
+            if chi0 < tol:
+                break
+        return vecs
+
+    def get_traj_from_sofb(self):
```
murilobalves, line 119: You are not using this function in the code, but I would like to remember you that SOFB returns the position in micrometers and the simulation runs using the position in meters. I don't know if the best to do is returning the values in meters or inserting a simple comment with the units is enough. Just avoiding silly problems with units.

apsuite/commissioning_scripts/inj_traj_fitting.py
```diff
@@ -0,0 +1,230 @@
...
+        x_ini, y_ini, xl_ini, yl_ini = trajx[0], trajy[0], 0, 0
+        de_ini = np.mean(trajx) / etax_ave
+        return np.array([x_ini, xl_ini, y_ini, yl_ini, de_ini])
+
+    def do_fitting(self, trajx, trajy, vec0=None, max_iter=5, tol=1e-6):
+        """."""
+        vec0 = vec0 if vec0 is not None else self.calc_init_vals(trajx, trajy)
+        res0 = self.calc_residue(vec0, trajx, trajy)
+        chi0 = self.calc_chisqr(res0)
+        vecs = [vec0, ]
```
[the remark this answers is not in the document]
fernandohds564: I thought about this, but this whole correction loop takes 1s to run and the visual inspection for the fitting quality is very direct. I decided to drop this idea because I think the importance of this data won't last long. Besides, with the resulting fitted vectors it would be easy to get `chi`in case the user wants it.

I could meet you half way and add a `full` flag to the input of the method in case the user wants the more detailed results. What do you think @murilobalves?

apsuite/commissioning_scripts/inj_traj_fitting.py
fernandohds564: Yeah, that's a good idea. I'll do it!

apsuite/commissioning_scripts/inj_traj_fitting.py
fernandohds564: meters!

I forgot the conversion factors, thanks!

Ps.: I don't use this method in the code, because It is supposed to the user call this method or the simulation method (`simulate_sofb`) in case it wants to fit a real or a simulated trajectory.
I'll add an usage example in the PR description.

apsuite/commissioning_scripts/inj_traj_fitting.py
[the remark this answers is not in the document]
murilobalves: Ok! The `chis` can be obtained from `vecs`, indeed. Since the correction loop is that fast, I think it's fine the way it's. Then, nevermind my comment.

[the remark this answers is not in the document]
fernandohds564: done!

apsuite/commissioning_scripts/inj_traj_fitting.py
fernandohds564: done!

apsuite/commissioning_scripts/inj_traj_fitting.py
fernandohds564: done!

PR #70, opened 2020-05-28
Add Tune Measurement class

xresende:


apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+from ..optimization import SimulAnneal as _SimulAnneal
+
+
+class FitTunes(_SimulAnneal):
+    """Tune Fit.
+
+    tunex = coeff1 * quad_parameter + offset1
+    tuney = coeff2 * quad_parameter + offset1
```
anacso17, line 14: Isn't it:  tuney = coeff2 * quad_parameter + **offset2** ?

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -0,0 +1,169 @@
...
+            fittune.fitting_plot()
+            fittune.niter = 1000
+            fittune.start()
+            fittune.fitting_plot()
+    """
+
+    def __init__(self, param, tune1, tune2):
+        """."""
+        self._param = _np.array(param)
+        self._tune1 = _np.array(tune1)
```
fernandohds564, line 32: it is better to use `_np.asarray(tune1)` here. It avoids unnecessary copies in case `tune1` already is an `numpy.ndarray`.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -0,0 +1,169 @@
...
+            title = 'Transverse Linear Coupling: {:.2f} %'.format(
+                position[-1] * 100)
+        prop_cycle = _plt.rcParams['axes.prop_cycle']
+        colors = prop_cycle.by_key()['color']
+        if fig is None:
+            fig = _plt.figure()
+        param, fittune1, fittune2 = FitTunes.calc_tunes(self._param, *position)
+
+        # plot meas data
+        _plt.plot(
```
fernandohds564, line 144: If you use `matplotlib.gridspec` you will be able to control the size and position of the `axes` inside the figure, improving visualization.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -0,0 +1,169 @@
+"""Coupling Measurement from Minimal Tune Separation."""
+
+import numpy as _np
+import matplotlib.pyplot as _plt
+
+
+from ..optimization import SimulAnneal as _SimulAnneal
+
+
+class FitTunes(_SimulAnneal):
```
fernandohds564, line 10: Wouldn't it be better to generalize this class to also perform the measurements?
Or to create a measurement class, in such a way that an object of this class is an attribute of the measurement class?
We already have a `BaseClass` for measurements in the `base.py` module of this subpackage.

apsuite/commissioning_scripts/meas_coupling_tune.py
xresende: It is! thanks!

apsuite/commissioning_scripts/meas_coupling_tune.py
xresende: I did not know this function. thanks!

apsuite/commissioning_scripts/meas_coupling_tune.py
xresende: Yes, I was going to do this (this is why I indicated that the PR is incomplete in the commit message). I wanted to upload this preliminary PR so you all could replicate the coupling fitting I did today.

I will study `base.py`

apsuite/commissioning_scripts/meas_coupling_tune.py
xresende: I implemented a _back-of-the-envelope_ plot method so that I could see the fitting results. You guys are much more familiar with _matplotlib_. Do you wan to commit a generalized plot method? (for example, one that allows users to choose where to plot).

apsuite/commissioning_scripts/meas_coupling_tune.py
xresende: on the other hand, since I need to sort tune arrays, I think it is safer to work with argument copies by invoking __np.array_. I think the additional memory/process overhead pays for reducing potential side effects of changing class user's data.

xresende: I changed the code so as not to incur in unnecessary copies, when possible. see what you think...

PR #71, opened 2020-05-28
New beta measurement 

murilobalves:
This new version does not cycle the trims before the measurement. 
The trim variation is always in the sense of increasing the power supply current (therefore for QFs the delta_kl is positive and for QDs it is negative).
After each trim variation there is a loop to recover the tune before the measurement.


murilobalves:
> @murilobalves , do you really think it is a good idea to remove cycling option from this class?
> In other words, is this solution we are using ideal?
> As I remember there is the inconvenient that we cannot save any configuration after measuring betas.



> @murilobalves , do you really think it is a good idea to remove cycling option from this class?
> In other words, is this solution we are using ideal?
> As I remember there is the inconvenient that we cannot save any configuration after measuring betas.

murilobalves:
> @murilobalves , do you really think it is a good idea to remove cycling option from this class?
> In other words, is this solution we are using ideal?
> As I remember there is the inconvenient that we cannot save any configuration after measuring betas.

Ops, unintentionally I closed the PR. Now I'm reopening it.

No problem, I can keep the cycling option.

Yes, there is this inconvenient, but it is not new. We still don't have a beta measurement that allows for saving a configuration after it happens.

PR #89, opened 2020-09-09
Delta kicks setting in respmat measurements

murilobalves:


fernandohds564 (approved):
Very nice! This was bothering me. Thanks for fixing it!.

apsuite/commissioning_scripts/measure_respmat.py
```diff
@@ -19,13 +19,19 @@ def __init__(self):
...
+        self.delta_kicky = 15  # [urad]
+        self.delta_rf = 80     # [Hz]
         self.respmat_name = ''
 
     def __str__(self):
         """."""
         dtmp = '{0:26s} = {1:9d}  {2:s}\n'.format
         stg = dtmp('nr_points_smooth', self.nr_points_smooth, '')
         stg += dtmp('corr_nr_iters', self.corr_nr_iters, '')
+        stg += dtmp('delta_kickx', self.delta_kickx, '[urad]')
```
fernandohds564, line 32: I think this is wrong. The kicks are floats, not integers. I think you have to create a string template for floats. I suggest using two decimal places.

fernandohds564 (changes requested):

apsuite/commissioning_scripts/measure_respmat.py
murilobalves: Indeed, thanks!

fernandohds564 (approved):

xresende (approved):

PR #97, opened 2020-10-05
Fix LOCO jacobian load

murilobalves:


anacso17 (approved):

apsuite/loco/main.py
```diff
@@ -200,41 +203,61 @@ def _handle_quad_fit_k(self, fname_jloco_k_quad):
                     jloco_k_quad_dict, 'quadrupole')
 
     def _convert_dict2array(self, jlocodict, magtype, is_normal=True):
-        jloco = []
+
+        if 'knobs_order' in jlocodict:
+            knobs_order = jlocodict['knobs_order']
```
fernandohds564, line 208: I'm kind of sad you didn't choose to go with the `OrderedDict` here. I think it would be more pythonic and of ease transition for the ordinary `dict` once we change python's version to `>=3.7`.
However, since this is a temporary fix, I don't think it will be a problem.

apsuite/loco/main.py
```diff
@@ -200,41 +203,61 @@ def _handle_quad_fit_k(self, fname_jloco_k_quad):
...
         quadfam = self.config.use_quad_families
         quadfam &= magtype == 'quadrupole'
         quadfam &= is_normal
         dipfam = self.config.use_dip_families
         dipfam &= magtype == 'dipole'
         dipfam &= is_normal
         if dipfam or quadfam:
+            jloco = []
             for quad in magstofit:
                 famcols = [
```
fernandohds564, line 244: I think it's `famcoils` here, right?

murilobalves: famcols is a short for family columns, the matrix columns that belong to the same family.

apsuite/loco/main.py
murilobalves: You are right, I'm sorry for disappointing you, but I intend to implement the more adequate solutions ASAP.

apsuite/loco/config.py
```diff
@@ -431,11 +432,28 @@ def update_sext_knobs(self):
...
+                raise Exception('invalid skew quadrupole name used to fit!')
+            else:
+                self.skew_quad_indices = []
+                for fam_name in self.skew_quadrupoles_to_fit:
+                    self.skew_quad_indices += self.respm.fam_data[
+                        fam_name]['index']
+                idx_all = _np.array(
+                    self.respm.fam_data['QS']['index']).flatten()
+                idx_sub = _np.array(self.skew_quad_indices).flatten()
+                self.skew_quad_indices = list(set(idx_sub) & set(idx_all))
```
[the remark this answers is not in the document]
murilobalves: Yes I am assuming no segmentation for QS. I am considering that the only segmented elements in the lattice are the dipoles and dealing with them accordingly. I concluded that since the re-factorization that we discussed is planned to happen ASAP and it will generalize the code and make it more robust against possible problems of that kind, if the question is not a bug, then it is not worth to spend much time on it. I truly expect that the re-factorization will happen sooner than significant modifications in the lattice, like changing the elements segmentation.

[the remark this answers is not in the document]
fernandohds564: ok!

apsuite/loco/main.py
fernandohds564: ok!

apsuite/loco/main.py
fernandohds564: kkk, my fault!

fernandohds564 (approved):

PR #105, opened 2021-01-11
Coupling measurement script

murilobalves:
Including MeasCoupling class that actually drives the measurement. Adapted from the jupyter notebook `meas_coupling` in machstudy_scrips of screens-iocs folder.

Change fitting method to use `scipy.optimization.least_squares`, which is faster and provides a better fit as compared to Simulated Annealing used previously.

Simulated Annealing:
`684 ms ± 4.43 ms per loop / residue: 1.4e-4`

Least Squares (Scipy):
`37.1 ms ± 241 µs per loop / residue: 7.4e-5`

### DIFF
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -17,8 +17,8 @@ class CoupParams():
         """."""
         self.nr_points = 1
         self.time_wait = 5  # s
-        self.neg_percent = 0.1/100
-        self.pos_percent = 0.1/100
+        self.neg_percent = 0.0/100
+        self.pos_percent = 0.0/100
 
     def __str__(self):
         """."""
@@ -88,15 +88,16 @@ class MeasCoupling(BaseClass):
         self.data['current'] = curr_vec
         self.data['tunes'] = tunes_vec
 
-    def process_and_plot_data(self):
+    def process_and_plot_data(self, coup_resolution=None, niter=2000):
         """."""
         _np.random.seed(seed=13101971)
         tune1, tune2 = self.data['tunes'].T
         curr = self.data['current']
 
-        FitTunes.COUPLING_RESOLUTION = 0.2/100
+        if coup_resolution:
+            FitTunes.COUPLING_RESOLUTION = coup_resolution
         fittune = FitTunes(param=curr, tune1=tune1, tune2=tune2)
-        fittune.niter = 2000
+        fittune.niter = niter
         fittune.start(print_flag=False)
         fittune.fitting_plot(oversampling=10, xlabel='QFB Variation [%]')
 
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -36,12 +36,13 @@ class MeasCoupling(BaseClass):
 
     QUAD_PVNAME = 'SI-FAM:PS-QFB'
 
-    def __init__(self):
+    def __init__(self, is_online=True):
         """."""
         super().__init__()
         self.params = CoupParams()
-        self.devices['quad'] = PowerSupply(MeasCoupling.QUAD_PVNAME)
-        self.devices['tune'] = Tune(Tune.DEVICES.SI)
+        if is_online:
+            self.devices['quad'] = PowerSupply(MeasCoupling.QUAD_PVNAME)
+            self.devices['tune'] = Tune(Tune.DEVICES.SI)
         self.data = dict()
         self._stopevt = _Event()
         self._thread = _Thread(target=self._do_meas, daemon=True)
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -199,7 +199,7 @@ class MeasCoupling(BaseClass):
             hspace=0.5, wspace=0.35)
         axi = _plt.subplot(grid[0, 0])
 
-        axi.set_xlabel(f'{anl["qname"]} Current [A]')
+        axi.set_xlabel(f'{self.data["qname"]} Current [A]')
         axi.set_ylabel('Transverse Tunes')
         fig.suptitle('Transverse Linear Coupling: ({:.2f} ± {:.2f}) %'.format(
             fit_vec[-1]*100, anl['fitting_error'][-1] * 100))
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -215,13 +215,11 @@ class MeasCoupling(BaseClass):
         coeff1, offset1, coeff2, offset2, coupling = params
         fx_ = coeff1 * curr + offset1
         fy_ = coeff2 * curr + offset2
-        tune1, tune2 = _np.zeros(curr.shape), _np.zeros(curr.shape)
-        for idx, _ in enumerate(curr):
-            mat = _np.array([[fx_[idx], coupling/2], [coupling/2, fy_[idx]]])
-            eigvalues, _ = _np.linalg.eig(mat)
-            tune1[idx], tune2[idx] = eigvalues
-            if tune1[idx] <= tune2[idx]:
-                tune1[idx], tune2[idx] = tune2[idx], tune1[idx]
+        coupvec = _np.ones(curr.size) * coupling/2
+        mat = _np.array([[fx_, coupvec], [coupvec, fy_]]).T
+        tune1, tune2 = _np.linalg.eigvalsh(mat).T
+        sel = tune1 <= tune2
+        tune1[sel], tune2[sel] = tune2[sel], tune1[sel]
         return tune1, tune2
 
     @staticmethod
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -119,7 +119,7 @@ class MeasCoupling(BaseClass):
         ini_param = self._calc_init_parms(qcurr, tune1, tune2)
         # least squares using Levenberg-Marquardt minimization algorithm
         fit_param = least_squares(
-            fun=MeasCoupling._err_func, x0=ini_param,
+            fun=self._err_func, x0=ini_param,
             args=(qcurr, tune1, tune2), method='lm')
 
         self.analysis['qcurr'] = qcurr
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -116,6 +116,15 @@ class MeasCoupling(BaseClass):
         """Wait for measurement to finish."""
         self._finished.wait(timeout=timeout)
 
+    def load_and_apply_old_data(self, fname):
+        """."""
+        data = self.load_data(fname)
+        if 'data' in data:
+            self.load_and_apply(fname)
+            return
+        data['timestamp'] = _os.path.getmtime(fname)
+        self.data = data
+
     def _do_meas(self):
         if not self.isonline:
             _log.error(
--- a/apsuite/commissioning_scripts/meas_coupling_tune.py
+++ b/apsuite/commissioning_scripts/meas_coupling_tune.py
@@ -167,17 +167,19 @@ class MeasCoupling(BaseClass):
 
     def process_data(self):
         """."""
+        self.analysis = dict()
         qcurr, tune1, tune2 = self._filter_data()
+        self.analysis['qcurr'] = qcurr
+        self.analysis['tune1'] = tune1
+        self.analysis['tune2'] = tune2
+
         ini_param = self._calc_init_parms(qcurr, tune1, tune2)
+        self.analysis['initial_param'] = ini_param
+
         # least squares using Levenberg-Marquardt minimization algorithm
         fit_param = least_squares(
             fun=self._err_func, x0=ini_param,
             args=(qcurr, tune1, tune2), method='lm')
-
-        self.analysis['qcurr'] = qcurr
-        self.analysis['tune1'] = tune1
-        self.analysis['tune2'] = tune2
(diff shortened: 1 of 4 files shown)


anacso17 (approved):

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -1,10 +1,106 @@
...
+        _np.random.seed(seed=13101971)
+        tune1, tune2 = self.data['tunes'].T
+        curr = self.data['current']
+
+        if coup_resolution:
+            FitTunes.COUPLING_RESOLUTION = coup_resolution
+        fittune = FitTunes(param=curr, tune1=tune1, tune2=tune2)
+        fittune.niter = niter
+        fittune.start(print_flag=False)
+        fittune.fitting_plot(oversampling=10, xlabel='QFB Variation [%]')
```
fernandohds564, line 103: It would be nice to separate data processing and plot. I think the class FitTunes should be incorporated to this one. The data resulted from the analysis should be placed in the analysis dictionary, as it is done in other measurement classes, such as BBA, InjTrajFittting...

fernandohds564: maybe, instead of the simulated annealing, we could use the method `curve_fit` from scipy. What do you think?

fernandohds564 (changes requested):

apsuite/commissioning_scripts/meas_coupling_tune.py
murilobalves: Ok, I'll study this.

xresende (approved):

apsuite/commissioning_scripts/meas_coupling_tune.py
murilobalves: @fernandohds564 done! please take a look.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -89,225 +102,144 @@ def _do_meas(self):
...
+        return _np.sqrt(_np.diag(pcov))
+
+    # static methods
+    @staticmethod
+    def _get_normal_modes(params, curr):
+        coeff1, offset1, coeff2, offset2, coupling = params
+        fx_ = coeff1 * curr + offset1
+        fy_ = coeff2 * curr + offset2
+        tune1, tune2 = _np.zeros(curr.shape), _np.zeros(curr.shape)
+        for idx, _ in enumerate(curr):
```
fernandohds564, line 219: I think you don't need this loop here. You can build a 3D array of shape `(n, 2, 2)` and call the method `eig` for it. I think this way it will be faster than the loop.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -89,225 +102,144 @@ def _do_meas(self):
...
+            mat = _np.array([[fx_[idx], coupling/2], [coupling/2, fy_[idx]]])
+            eigvalues, _ = _np.linalg.eig(mat)
```
fernandohds564, line 221: Since we are dealing with a real symmetric matrix, I think you can use `eigh` instead of `eig` here. Actually, you can use `eigvalsh`, since we don't care for eigenvectors.

fernandohds564 (changes requested):
very nice @murilobalves ! There are a few comments in the code, please take a look.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -89,225 +102,144 @@ def _do_meas(self):
...
+        # based on fitting error calculation of scipy.optimization.curve_fit
+        # do Moore-Penrose inverse discarding zero singular values.
+        fit_params = self.analysis['fitted_param']
+        _, smat, vhmat = _np.linalg.svd(
+            fit_params['jac'], full_matrices=False)
+        thre = _np.finfo(float).eps * max(fit_params['jac'].shape)
+        thre *= smat[0]
+        smat = smat[smat > thre]
+        vhmat = vhmat[:smat.size]
+        pcov = _np.dot(vhmat.T / smat*smat, vhmat)
```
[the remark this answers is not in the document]
murilobalves: thanks!

apsuite/commissioning_scripts/meas_coupling_tune.py
murilobalves: Indeed, the loop was substituted by the 3D array approach.

apsuite/commissioning_scripts/meas_coupling_tune.py
murilobalves: Thanks for the advice, now the code uses `eigvalsh`!

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
+        qcurr, tune1, tune2 = anl['qcurr'], anl['tune1'], anl['tune2']
+
+        qcurr_interp = _np.interp(
+            _np.arange(0, len(qcurr), 1/oversampling),
+            _np.arange(0, len(qcurr)), qcurr)
+
+        fittune1, fittune2 = MeasCoupling._get_normal_modes(
+            params=fit_vec, curr=qcurr_interp)
 
+        if fig is None:
```
fernandohds564, line 137: I think the optional argument should be the `axes` object, not the figure. I recommend substitute the `_plt.` by the `axes.`

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
-        _plt.plot(param, fittune1, color=colors[1], label='fitting')
-        _plt.plot(param, fittune2, color=colors[1])
+        _plt.plot(qcurr_interp, fittune1, color='tab:gray', label='fitting')
+        _plt.plot(qcurr_interp, fittune2, color='tab:gray')
         _plt.legend()
         _plt.xlabel(xlabel)
         _plt.ylabel('Transverse Tunes')
         _plt.title(title)
-        _plt.grid()
+        _plt.grid(ls='--', alpha=0.5)
```
fernandohds564, line 156: I think these general visualization parameters, such as this one and the tight_layout do not belong here. I think it is better to let these configs for the user.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
+        _plt.tight_layout(True)
+        if save:
+            if fname is None:
+                date_string = _time.strftime("%Y-%m-%d-%H:%M")
+                fname = 'coupling_fitting_{}.png'.format(date_string)
+            _plt.savefig(fname, format='png', dpi=300)
         _plt.show()
```
fernandohds564, line 163: I think it is a good practice to return the `Figure` and the `Axes` objects for the user to post configure the graph.
I think it is better not to call `_plt.show()` either.

fernandohds564 (changes requested):
@murilobalves , have you tested this vectorized way of calculating the eigenvalues? It seems so simple now that I wonder if that's only that :smile: !

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -1,14 +1,38 @@
...
+from .base import BaseClass
 
-from ..optimization import SimulAnneal as _SimulAnneal
 
+class CouplingParams():
+    """."""
+
+    def __init__(self):
+        """."""
+        self.nr_points = 1
```
fernandohds564, line 18: I think the default value of this attribute should be more appropriate, @murilobalves.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
+        tunes = self.devices['tune']
+
+        curr0 = quad.current
+        curr_vec = curr0 * _np.linspace(
+            1-self.params.neg_percent,
+            1+self.params.pos_percent,
+            self.params.nr_points)
+        tunes_vec = _np.zeros((len(curr_vec), 2))
+
+        print('QFB Current:')
```
fernandohds564, line 91: I think this hard-coded quadrupole name should be replaced by the one from the `quad` object.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
+        for idx, curr in enumerate(curr_vec):
+            quad.current = curr
+            _time.sleep(self.params.time_wait)
+            tunes_vec[idx, :] = tunes.tunex, tunes.tuney
+            print('   {:.6f} A ({:+.3f} %), nux = {:.4f}, nuy = {:.4f}'.format(
```
fernandohds564, line 96: I think it is better to specify the sizes of the printed version of the variables, so all text is well aligned.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +41,203 @@ class FitTunes(_SimulAnneal):
...
+                curr, (curr/curr0-1)*100),
+                tunes_vec[idx, 0], tunes_vec[idx, 1])
+        print('Finished!')
+        quad.current = curr0
+        self.data['qname'] = quad.devname
+        self.data['current'] = curr_vec
+        self.data['tunes'] = tunes_vec
```
fernandohds564, line 103: I think it is important to save the time when the measurement was performed, so we can recover additional information via archiver later on.

fernandohds564 (changes requested):

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -78,6 +82,10 @@ def ismeasuring(self):
         return self._thread.is_alive()
 
     def _do_meas(self):
+        if self.devices['quad'].devname != self.params.quadfam_name:
```
fernandohds564, line 85: This will always be `False`. The attribute `devname` is the complete name of the device. In order to get what you want here you need to compare the variable `quadfam_name` with the attribute `dev` of the `SiriusPVName` object.

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -134,33 +143,36 @@ def plot_fitting(
...
         if save:
             if fname is None:
                 date_string = _time.strftime("%Y-%m-%d-%H:%M")
                 fname = 'coupling_fitting_{}.png'.format(date_string)
-            _plt.savefig(fname, format='png', dpi=300)
-        _plt.show()
+            fig.savefig(fname, format='png', dpi=300)
+        return fig, axi
 
     def _filter_data(self, coup_resolution=None):
```
fernandohds564, line 177: Wouldn't be better to remove this argument at all?

fernandohds564 (changes requested):

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -96,7 +96,7 @@ def _do_meas(self):
             self.params.nr_points)
         tunes_vec = _np.zeros((len(curr_vec), 2))
 
-        print('{:s} current:'.format(quad.devname))
+        print('{:s} current:'.format(quad.devname.dev))
```
fernandohds564, line 99: Here I think it is better to use the entire `devname`.

fernandohds564 (approved):

anacso17 (approved):

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +47,207 @@ class FitTunes(_SimulAnneal):
...
-        # NOTE: do error estimate properly!
-        coup_error = self.calc_obj_fun()
+    def ismeasuring(self):
+        """."""
+        return self._thread.is_alive()
+
+    def _do_meas(self):
+        if self.devices['quad'].devname.dev != self.params.quadfam_name:
+            self.devices['quad'] = PowerSupply(
+                'SI-Fam:PS-' + self.params.quadfam_name)
```
fernandohds564, line 87: @murilobalves, I think there is a `.wait_for_connection` missing below this line, outside the `if` statement to guarantee the PVs are connected.

fernandohds564 (changes requested):

apsuite/commissioning_scripts/meas_coupling_tune.py
```diff
@@ -17,201 +47,207 @@ class FitTunes(_SimulAnneal):
...
+            s_sq = cost / (ysize - popt.size)
+            pcov = pcov * s_sq
+        else:
+            pcov.fill(_np.nan)
+            print('# of fitting parameters larger than # of data points!')
+        return _np.sqrt(_np.diag(pcov))
+
+    # static methods
+    @staticmethod
+    def _get_normal_modes(params, curr):
```
fernandohds564, line 223: This method should be part of the API (without the leading `_`). I will use it in the `MeasCouplingWindow`

fernandohds564 (changes requested):

apsuite/commissioning_scripts/measure_beta.py
```diff
@@ -188,7 +188,7 @@ def anly_method(self, value):
 
     def start(self):
         """."""
-        if self._thread.is_alive():
+        if not self.is_online or self._thread.is_alive():
```
fernandohds564, line 191: @murilobalves, I am using the name `isonline` istead of `is_online` in the coupling class. I did this because of the `ismeasuring` property. Which one do you prefer?

murilobalves: Ok, let's keep this pattern

fernandohds564 (approved):

fernandohds564:
@xresende, I think this PR is finally ready for revision. Could you please take a look at it?

xresende (approved):

PR #118, opened 2021-02-22
Fix print in apply_strengths

murilobalves:


anacso17 (approved):

apsuite/commissioning_scripts/apply_strengths.py
```diff
@@ -306,7 +310,7 @@ def apply_model_strengths(
                 magidx = fam['QN']['index']
                 # NOTE: This is incorrect for magnets with more than one
```
fernandohds564, line 311: @murilobalves, I think this `NOTE` would be better placed inside the method `_get_idx`, where the algorithm is really placed.

fernandohds564 (approved):

apsuite/commissioning_scripts/apply_strengths.py
murilobalves: Indeed. Done!

xresende (approved):

PR #131, opened 2021-03-19
Parallelize LOCO jacobian

murilobalves:
thanks to @fernandohds564 for breaking the inertia and implementing the template for parallelization.


anacso17 (approved):

apsuite/loco/utils.py
```diff
@@ -243,110 +244,148 @@ def jloco_calc_linear(config, matrix):
...
-            _pyaccel.lattice.get_attribute(model, 'KL', sn_indices))
-        set_quad_kdelta = LOCOUtils.set_quadmag_kdelta
+        if config.use_sext_families:
+            sindices = []
+            for fam_name in config.famname_sextset:
+                sindices.append(config.respm.fam_data[fam_name]['index'])
+        else:
+            sindices = config.respm.fam_data['SN']['index']
+        magtype = 'sextupole'
+        if config.parallel is not None:
```
fernandohds564, line 323: In order to cleanup the code you could put this if inside the parallel base method.

I also think the variable parallel could be False, True or an integer specifying the number of processes to use. In case it is True the code decides how many processes based on the number of cores of the machine.
For an example, take a look at my PR on pyaccel.

fernandohds564 (commented):
Very nice @murilobalves !

I left one comment in the code.

fernandohds564 (approved):

xresende (approved):

PR #141, opened 2021-07-06
MNT: update functions name change from siriuspy.

fernandohds564:


anacso17 (approved):

murilobalves (approved):

xresende (commented):
I just realised now that a PR in siriuspy changes the method names in RFCav device: from 'cmd_set_phase' to 'set_phase'. But other devices kept using the prefix 'cmd_'. I think we should try to keep consistent notation. I think prefixes are useful in order to,  in completion editing actions, group methods  that share characteristics. In this particular case, methods that evaluate to device commands. I would keep 'cmd_' but, if not, I think we should then rename the corresponding methods from other device objects.

xresende (changes requested):
I just realized now that a PR in siriuspy changes the method names in RFCav device: from 'cmd_set_phase' to 'set_phase'. But other devices kept using the prefix 'cmd_'. I think we should try to keep consistent notation. I think prefixes are useful in order to, in completion editing actions, group methods that share characteristics. In this particular case, methods that evaluate to device commands. I would keep 'cmd_' but, if not, I think we should then rename the corresponding methods from other device objects.

fernandohds564:
@xresende , the `cmd_` prefix is used for methods that do not take inputs (apart from `timeout`). They are commands that always execute the same task. I removed the `cmd_` from these methods because the are setpoint methods and do take inputs.

xresende:
@fernandohds564 , https://github.com/lnls-sirius/dev-packages/pull/709#issuecomment-875828427

xresende (approved):

PR #143, opened 2021-10-19
Touschek lifetime measurement

murilobalves:


apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -468,6 +389,301 @@ def plot_current_decay(self, fname=None, title=None):
...
+        excx0 = self.pvs['exc_tunex_sts'].value
+        excy0 = self.pvs['exc_tuney_sts'].value
+
+        self.devices['trigger'].source = 'Study'
+        self.devices['event'].mode = 'Continuous'
+        self.devices['evg'].cmd_update_events()
+
+        self.turn_off_bpms_auto_monitor()
+        bpm.acq_nrsamples_pre = parms.acq_nrsamples_pre
+        bpm.acq_nrsamples_post = parms.acq_nrsamples_post
```
fernandohds564, line 610: I think we could set the property `bpm.acq_repeat` to `"normal"`. This way the BPM will not rearm the trigger right after the acquisition, which will allow us to set the mask without the need to `abort` the acquisition.

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -468,6 +389,301 @@ def plot_current_decay(self, fname=None, title=None):
...
+        meas['tunex'] = tune.tunex
+        meas['tuney'] = tune.tuney
+        _time.sleep(1)
+        self.pvs['exc_tunex_sel'].value = 0
+        self.pvs['exc_tuney_sel'].value = 0
+
+        while idx < maxidx and not self._stopevt.is_set():
+            bpm.tbt_mask_beg = parms.mask_beg_bunch_a
+            bpm.tbt_mask_end = parms.mask_end_bunch_a
+            _time.sleep(parms.wait_mask)
```
fernandohds564, line 635: Here we should `atart` the acquisition, send the `event` and wait for the BPM `acq_status` property go to `"idle"`. Then wait a little longer for all the PVs we need to be updated. Maybe this last _waiting_ should be setup via callback of the PVs. (we can chat if this is not clear).

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -468,6 +389,301 @@ def plot_current_decay(self, fname=None, title=None):
...
+            meas['sum_a'].append(bpm.mt_possum.mean())
```
fernandohds564, line 636: We should check whether the possum has `NaNs` and discard the measurement. Or we could use the `nanmean` function. Also, it would be nice to keep track of how many numbers were `NaN`s, maybe by some print message, or keeping their quantity in an array...

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -468,6 +389,301 @@ def plot_current_decay(self, fname=None, title=None):
...
+            meas['tim_a'].append(_time.time())
+
+            meas['current'].append(curr.current)
+            meas['rf_voltage'].append(rfcav.dev_cavmon.gap_voltage)
+            meas['avg_pressure'].append(press.value)
+
+            bpm.tbt_mask_beg = parms.mask_beg_bunch_b
+            bpm.tbt_mask_end = parms.mask_end_bunch_b
+            _time.sleep(parms.wait_mask)
```
fernandohds564, line 645: same as above, we should `start` the acquisition, send the `event`....

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -50,253 +69,158 @@ def __str__(self):
...
+                mstr += (f'\n{name:<20s}: '
+                    'rb {bpm.rffe_att:.0f} != sp {val_old[name]:.0f}')
+        if len(mstr) == len('except:'):
+            print('RFFE attenuation set confirmed in all BPMs.')
+        else:
+            print('RFFE attenuation set confirmed in all BPMs, ' + mstr)
+        
 
-    def switch_to_single_bunch(self):
+    def turn_off_bpms_auto_monitor(self):
```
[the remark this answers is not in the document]
fernandohds564: which one? I don't see anything related.

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -50,253 +69,158 @@ def __str__(self):
...
-        bpmsnames = BPMSearch.get_names(si_bpm_filter)
-        bpms = {name: BPM(name) for name in bpmsnames}
-        _ = [bpm.wait_for_connection() for bpm in bpms.values()]
-        return bpms
-
-    def set_si_bpms_attenuation(self, bpms=None, value_att=RFFEAttSB):
+            self.devices['rfcav'] = RFCav(RFCav.DEVICES.SI)
+            self.devices['tune'] = Tune(Tune.DEVICES.SI)
+            self.pvs['avg_pressure'] = PV(MeasTouschekLifetime.AVG_PRESSURE_PV)
+            self.pvs['exc_tunex_sts'] = PV(
```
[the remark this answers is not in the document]
xresende: see https://github.com/lnls-sirius/dev-packages/pull/747/files

apsuite/commisslib/meas_touschek_lifetime.py
```diff
@@ -50,253 +69,163 @@ def __str__(self):
...
+    def create_bpms(self):
         """."""
-        if bpms is None:
-            bpms = self._create_si_bpms()
-
-        val_old = dict()
-        for name, bpm in bpms.items():
-            val_old[name] = bpm.rffe_att
+        bpmsnames = BPMSearch.get_names({'sec': 'SI', 'dev': 'BPM'})
+        cbn = self.params.bpmname  # chosen bpm name
```
fernandohds564, line 114: I think we don't need to separate apart the chosen BPM here. All BPMs should be equal up to the point when the measurement starts, in case the user decided to change the chosen BPM after the object creation.

fernandohds564 (changes requested):

PR #161, opened 2021-12-14
TurnOff CHs subset with stored beam 

murilobalves:
Ramp down and turn off CHs subset while keeping stored beam.

Perform the process of ramping down the kicks for a list of specific horizontal correctors until zero current and then turn off its power supplies, while correcting COD and tunes.

At each iteration:
  - calc. expected COD due to reduction in CHs subset
  - calc. kicks in remaining CHs to compensate the expected COD
  - apply kicks that reduces the CHs subset and also compensate COD
  - check betatron tunes drifts and if so, correct the tunes
  - apply a few orbit corrections to minimize the residual COD

Kick variations during the CHs ramping down are limited to produce a maximum orbit distortion given by `params.max_delta_orbit`. The actual COD after each kick reduction is always smaller than `params.max_delta_orbit` since the compensation with the remaining CHs is applied at the same time. Therefore `params.max_delta_orbit` admits some flexibility.

Once CHs kicks reach zero, then its power supplies are turned off.
If a beam dump occurs during the process, it will be aborted.

A method to reverse the process is also available in this class: `restore_initial_state()`.

### DIFF
--- a/apsuite/commisslib/turnoff_correctors.py
+++ b/apsuite/commisslib/turnoff_correctors.py
@@ -50,6 +50,8 @@ class TurnOffCorrParams(_ParamsBaseClass):
 class TurnOffCorr(_ThreadBaseClass):
     """."""
 
+    MIN_CURRENT = 0.1
+
     def __init__(self, params=None, isonline=True):
         """."""
         params = TurnOffCorrParams() if params is None else params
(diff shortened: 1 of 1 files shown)


anacso17 (approved):

apsuite/commisslib/turnoff_correctors.py
```diff
@@ -0,0 +1,347 @@
...
+        self.devices['tunecorr'] = TuneCorr(TuneCorr.DEVICES.SI)
+        self.devices['sofb'] = SOFB(SOFB.DEVICES.SI)
+        self.devices['sibpms'] = FamBPMs(FamBPMs.DEVICES.SI)
+        self.devices['event'] = Event('Study')
+        self.devices['rfgen'] = RFGen()
+
+    def get_orbit_data(self):
+        """Get orbit data from BPMs in Monit1 acquisiton rate.
+
+        BPMs must be configure to listen to Study event and the Study event
```
vellosok75, line 132: Typo: *configure* -> configured

fernandohds564 (approved):

apsuite/commisslib/turnoff_correctors.py
murilobalves: Thanks @vellosok75. Fixed.

xresende (approved):

jquentino (approved):

vellosok75 (approved):

PR #178, opened 2022-03-10
SCRIPTS.ENH: Adds script that simulates the nlk field

jquentino:
This script contains two classes, one to generate the field of an infinite wire and another to simulate the NLK field, constructed by using the previous wires objects. 
The below graphs show a comparison between the data from an excel sheet with the NLK nominal field and this simulated model. In addition, it's included plots of the bidimensional fields generated by a Wire object and an NLK object.

**1. Comparison between the field generated by the code and the field generated by excel. I also add a curve including the NLK field with random errors in the positions of the wires, sampled from a normal distribution.**

```python
import numpy as np
import matplotlib.pyplot as plt
from generic_scripts.nlk import Wire, NLK
import pandas as pd

# Pandas for simplifying the importation of data with ',' as decimal separator
data = pd.read_csv(
    'auxiliar_scripts/nlk_profile.txt', delimiter='\t', decimal=',', header=None, dtype='float')
data_nmp = data.to_numpy()
x_proj, field_proj = data_nmp[:, 0], data_nmp[:, 1]

nlk = NLK()
x, field = nlk.get_vertical_magnetic_field()

default_positions = nlk.positions
error = np.random.normal(0, 2e-4, default_positions.shape)
nlk.positions = nlk.positions + error
_, field_w_errors = nlk.get_vertical_magnetic_field()

fig, ax = plt.subplots()
ax.plot(x_proj, field_proj/10000, label='Excel field', lw=6, c = 'tab:blue')
ax.plot(x*1e3, field, label = 'Code field', ls='-', lw=3 , c = 'tab:orange')
ax.plot(x*1e3, field_w_errors, label= "Code field with random errors", lw=3, ls='-.', c='tab:red')
ax.set_ylabel('B [T]')
ax.set_xlabel('x [mm]')
ax.set_title(r'NLK field')
plt.legend()
plt.savefig('figures/nlk/nlk_noerrors.jpeg', dpi=160)
plt.show()
``` 
nlk_noerrors

**2. Code to plot the field of a Wire object:**

```python
wire = Wire(current=100)

x_space = np.linspace(-2.5, 2.5, num=25)[None, :]*1e-3
y_space = np.linspace(-2.5, 2.5, num=25)[None, :]*1e-3

xmesh, ymesh = np.meshgrid(x_space, y_space)

xrav = xmesh.ravel()[None, :]
yrav = ymesh.ravel()[None, :]

xpos = wire.position[0]
ypos = wire.position[1]

wire_radius = 0.59e-3
mask = (xrav - xpos)**2 + (yrav - ypos)**2 > (wire_radius)**2

r = np.vstack([xrav[mask], yrav[mask]])
wire_field = wire.calc_magnetic_field(r)

mag = np.hypot(wire_field[0], wire_field[1])
u_norm, v_norm =  wire_field[0, :]/mag, wire_field[1, :]/mag

plt.quiver(r[0,:]*1e3, r[1,:]*1e3, u_norm, v_norm, mag, units='width', pivot='tail',cmap='viridis')
plt.title("Field of a wire at (0,0), current=100 A", fontsize=18)
plt.xlabel('x [mm]')
plt.ylabel('y [mm]')
plt.colorbar(label='B [T]')
plt.savefig(save_path+'wire_field.png', transparent=False, facecolor='white', dpi=150)
```
wire_field

**3. Code to plot the NLK field**

```python
x_space = np.linspace(-12, 12, num=30)[None, :]*1e-3
y_space = np.linspace(-8, 8, num=30)[None, :]*1e-3
xmesh, ymesh = np.meshgrid(x_space, y_space)
xrav = xmesh.ravel()[None, :]
yrav = ymesh.ravel()[None, :]
xpos = nlk.positions[0]
ypos = nlk.positions[1]
mask = np.ones(xrav.shape, dtype=bool)
for i in range(xpos.size):
    mask *= (xrav - xpos[i])**2 + (yrav - ypos[i])**2 > (wire_radius)**2
r = np.vstack([xrav[mask], yrav[mask]])
r.shape
nlk_field = nlk.calc_magnetic_field(r)
mag = np.hypot(nlk_field[0], nlk_field[1])
u_norm, v_norm =  nlk_field[0, :]/mag, nlk_field[1, :]/mag

plt.quiver(r[0,:]*1e3, r[1,:]*1e3, u_norm, v_norm, mag, units='width', pivot='tail',cmap='viridis')
plt.title("NLK bidimensional field, current=1850 A", fontsize=18)
plt.xlabel('x [mm]')
plt.ylabel('y [mm]')
plt.colorbar(label='B [T]')
plt.savefig(save_path+'nlk_field.png', transparent=False, facecolor='white', dpi=150)
```
nlk_field

**4. The same field, but with magnitudes colors in logscale.**

plt.quiver(r[0,:]*1e3, r[1,:]*1e3, u_norm, v_norm, np.log(mag), units='width', pivot='tail',cmap='viridis')
plt.title("NLK bidimensional field, current=1850 A", fontsize=18)
plt.xlabel('x [mm]')
plt.ylabel('y [mm]')
plt.colorbar(label='log(B [T])')
plt.savefig(save_path+'nlk_field_log.png', transparent=False, facecolor='white', dpi=150)

nlk_field_log


generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+from mathphys.beam_optics import beam_rigidity
+
+
+class Wire:
+    """."""
+    def __init__(self, position, current):
+        self.pos = position  # Column vector
+        self.curr = current
+
+    def Bfield(self, x, y):
```
fernandohds564, line 13: Use lower case here. Besides, since this is a method, use one of the prefixes: `get_`, `calc_`, or similar.
Maybe a good name would be: `calc_magnetic_field()`

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        mu0 = vacuum_permeability
+        r_w = self.pos  # Wire positions
+        r = _np.array([x, y])[:, None]  # Target position
+        rc = r - r_w  # Cursive r
+        theta = _np.arctan2(rc[1], rc[0])[0]
+        theta_vec = _np.array([-_np.sin(theta), _np.cos(theta)])[:, None]
+        B = mu0 * self.curr/(2*_np.pi*_np.linalg.norm(rc))*theta_vec + r_w
+        return B
+
+    def include_pos_error(self, sigma):
```
fernandohds564, line 23: I don't think it is a good idea to create random number at this level.
I don't think this functionality of insert errors in this class is needed, because it constrain the way they are created.
The final user should create the errors at the top-level script and set the positions of the wires.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
+import numpy as _np
+import matplotlib.pyplot as _plt
+from mathphys.constants import vacuum_permeability
+from mathphys.beam_optics import beam_rigidity
+
+
+class Wire:
+    """."""
+    def __init__(self, position, current):
```
fernandohds564, line 9: include docsctrings in all methods to explain them and the type and format of the inputs.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        self.pos = position  # Column vector
+        self.curr = current
+
+    def Bfield(self, x, y):
+        mu0 = vacuum_permeability
+        r_w = self.pos  # Wire positions
```
fernandohds564, line 15: This could be a source of errors. You assume here that `self.pos` is a column vector, but the setter does not force this shape. I think you should do: `r_w = self._pos[:, None]` here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        r = _np.array([x, y])[:, None]  # Target position
```
fernandohds564, line 16: it would be better to change the input of this method. you should accept as input the vector(s) where the field must be calculated, not the `x` and `y` positions separately.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._curr
+
+    @curr.setter
+    def curr(self, current):
+        self._curr = current
+
+
+class NLK:
+    """."""
+    def __init__(self, curr=1850, errors=False, sigma_errors=1e-5):
```
fernandohds564, line 46: you don't need to add options to insert errors in this class. The user should add errors setting the position of the wires.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        # Setting wire positions
+        wire_positions = []
+        wire_positions.append(_np.array([7.0,  5.21])[:, None]*1e-3)
```
fernandohds564, line 49: The position of the wires could be an optional input to this class. You could create a class variable keeping the default positions.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        wire_positions.append(_np.array([10.,  5.85])[:, None]*1e-3)
+        wire_positions.append(_np.array([-7.,  5.21])[:, None]*1e-3)
+        wire_positions.append(_np.array([-10,  5.85])[:, None]*1e-3)
+        wire_positions.append(_np.array([7.0, -5.21])[:, None]*1e-3)
+        wire_positions.append(_np.array([10., -5.85])[:, None]*1e-3)
+        wire_positions.append(_np.array([-7., -5.21])[:, None]*1e-3)
+        wire_positions.append(_np.array([-10, -5.85])[:, None]*1e-3)
```
fernandohds564, line 56: make a loop here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        self._curr = current
+
+
+class NLK:
+    """."""
+    def __init__(self, curr=1850, errors=False, sigma_errors=1e-5):
+        # Setting wire positions
+        wire_positions = []
+        wire_positions.append(_np.array([7.0,  5.21])[:, None]*1e-3)
+        wire_positions.append(_np.array([10.,  5.85])[:, None]*1e-3)
```
fernandohds564, line 50: do not treat the wire positions as a list of `numpy.ndarray`s, but as a 2d `numpy.ndarray.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+            if errors:
+                wire.include_pos_error(sigma=sigma_errors)
+            self._wires.append(wire)
+
+    @property
+    def wires(self):
+        return self._wires
+
+    @wires.setter
+    def wires(self, wires_objects):
```
fernandohds564, line 83: I think the `wires` property should be readable only.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+    @property
+    def currents(self):
+        "Return currents of all wires"
+        return self._currents
+
+    @currents.setter
+    def currents(self, wires_currents):
+        self._currents = wires_currents
+
+    def Bfield(self, x, y):
```
fernandohds564, line 104: change the name and the input type of this method.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        field = _np.zeros([2, 1])
+        for wire in self.wires:
+            field += wire.Bfield(x, y)
+        return field
+
+    def include_pos_error(self, sigma_errors):
```
fernandohds564, line 110: remove methods that create errors here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        for wire in self.wires:
+            wire.include_pos_error(sigma_errors)
+
+    def nlk_profile(self):
```
fernandohds564, line 114: I think the name of this method could be different. Maybe `get_magnetic_field_on_axis` or similar.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        "NLK vertical field at y=0 for x ∈ [-12, 12] mm."
+        x_space = _np.linspace(-12, 12)*1e-3
+        fieldy = _np.array([self.Bfield(x, 0)[1] for x in x_space])
+        return x_space, fieldy
+
+
+def polyfit(x, y, n):
```
fernandohds564, line 121: why not use `numpy.polynominal.polinomial.polyfit` here?

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+    """Generates the nlk integrated polynom_b and its horizontal
+    kick."""
+
+    if fit_monomials is None:
+        fit_monomials = _np.arange(10, dtype=int)
+    if strength is None:
+        strength = 0.27358145
+
+    nlk = NLK(errors=errors, sigma_errors=sigma_errors)
+    x, B = nlk.nlk_profile()
```
fernandohds564, line 147: use lower case variable names.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        _plt.figure()
+        _plt.scatter(1e3*(x-r0), 1e3*kickx, label="data points")
+        _plt.plot(
+            1e3*(x-r0), 1e3*fit_kickx, c=[0, 0.6, 0], label="fitted curve")
+        _plt.xlabel('X [mm]')
+        _plt.ylabel("Kick @ 3 GeV [mrad]")
+        _plt.title("NLK Profile")
+        _plt.legend()
+
+    LPolyB = _np.zeros([1, 1 + _np.max(fit_monomials)])
```
fernandohds564, line 165: use lower case variable names. this format for the `polynom_b` is not intuitive.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+    @curr.setter
+    def curr(self, current):
+        self._curr = current
+
+
+class NLK:
+    """."""
+    def __init__(self, curr=1850, errors=False, sigma_errors=1e-5):
+        # Setting wire positions
+        wire_positions = []
```
murilobalves, line 48: Maybe take `wire_positions` as an input (a list) and make this default list in case of nothing is passed by the user. 
Actually I think it would be nice to set different values of currents for each wire. Maybe this class could accept a dict like `wires_info` containing the lists `wire_positions` and the corresponding `wire_currents`.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+    if fit_monomials is None:
+        fit_monomials = _np.arange(10, dtype=int)
+    if strength is None:
+        strength = 0.27358145
+
+    nlk = NLK(errors=errors, sigma_errors=sigma_errors)
+    x, B = nlk.nlk_profile()
+
+    brho, *_ = beam_rigidity(energy=3)  # Energy in GeV
+    integ_field = strength * B
```
fernandohds564, line 150: the integrated field is the length of the magnet times the field.

fernandohds564 (changes requested):

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        for wire in self.wires:
+            field += wire.Bfield(x, y)
+        return field
+
+    def include_pos_error(self, sigma_errors):
+        for wire in self.wires:
+            wire.include_pos_error(sigma_errors)
+
+    def nlk_profile(self):
+        "NLK vertical field at y=0 for x ∈ [-12, 12] mm."
```
murilobalves, line 115: Maybe take an input `y` with default `y=0` to specify the plane to be analyzed. Also it would be good to have a method to return the horizontal field at a fixed x. This would be useful to analyze these field components in the case of errors in wires.

generic_scripts/nlk.py
murilobalves, line 121: ok, it is nice to implement a polynomial fitting by yourself, but I think `numpy.polynomial.polynomial.polyfit` optimized.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+def si_nlk_kick(
+        strength=None, fit_monomials=None, plot_flag=False,
+        r0=0.0, errors=False, sigma_errors=1e-5):
+    """Generates the nlk integrated polynom_b and its horizontal
+    kick."""
+
+    if fit_monomials is None:
+        fit_monomials = _np.arange(10, dtype=int)
+    if strength is None:
+        strength = 0.27358145
```
murilobalves, line 144: which unit is this?

generic_scripts/nlk.py
murilobalves, line 150: maybe specify that this in in units of [T.m]

murilobalves (changes requested):

generic_scripts/nlk.py
fernandohds564: I think that in the initialization of the object it is not so important to set each current individually. Since we have setters for the current of the wires. Besides, not using a dictionary simplifies the use.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+    def wires(self, wires_objects):
+        self._wires = wires_objects
+
+    @property
+    def positions(self):
+        "Return positions of all wires"
+        return self._positions
+
+    @positions.setter
+    def positions(self, wires_positions):
```
fernandohds564, line 92: this setter method should in fact set the positions of each wire.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._wires
+
+    @wires.setter
+    def wires(self, wires_objects):
+        self._wires = wires_objects
+
+    @property
+    def positions(self):
+        "Return positions of all wires"
+        return self._positions
```
fernandohds564, line 89: this getter should read the positions of each wire.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @positions.setter
+    def positions(self, wires_positions):
+        self._positions = wires_positions
+
+    @property
+    def currents(self):
+        "Return currents of all wires"
+        return self._currents
```
fernandohds564, line 98: same here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @currents.setter
+    def currents(self, wires_currents):
+        self._currents = wires_currents
```
fernandohds564, line 102: here too.

fernandohds564 (changes requested):

generic_scripts/nlk.py
jquentino: I think that would be better to follow the idea of forcing the shape in the setter and raise an exception if the array isn't of size 2.

generic_scripts/nlk.py
xresende: I like the concise format `  # [T.m]`, `  # [GeV]` (with bracketing as a default indication of inserted string as units)

generic_scripts/nlk.py
fernandohds564: I don't know about forcing the shape, but you could force the size of the first dimension to be 2.

generic_scripts/nlk.py
jquentino: Done!

generic_scripts/nlk.py
jquentino: Done!

generic_scripts/nlk.py
jquentino: I didn't know this function, only numpy.polyfit, but in the last one you can't pass the polynomial powers that were to be used in the fit, just the degree of the polynomial, so I decided not to use. 
On the other hand,  numpy.polynominal.polinomial.polyfit has this option, so I added it on the code in [5cf73b0](https://github.com/lnls-fac/apsuite/pull/178/commits/5cf73b038c593dfff71c4c69f9b05077809fd588). 
Thanks for the suggestion!

generic_scripts/nlk.py
jquentino: Yes, I forgot to multiply by L. Solved in [polynomial](https://www.google.com/search?client=ubuntu&hs=Bp&channel=fs&sxsrf=APq-WBspzVh9Igo6N_7XNJCLH42uHyqoVA:1647349822563&q=polynomial&spell=1&sa=X&ved=2ahUKEwiZifqnmMj2AhUkq5UCHYaCBTIQkeECKAB6BAgCEDU).

generic_scripts/nlk.py
jquentino: Adimensional, works like a pulse amplifier.

generic_scripts/nlk.py
jquentino: Bug solved in [5da7e34](https://github.com/lnls-fac/apsuite/pull/178/commits/5da7e34230705947c7f1394521605248ec723351). But isn't possible yet to set the wire positions individually.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+    @property
+    def currents(self):
+        """Currents of the NLK wires in [A].
+
+        Returns
+        -------
+        numpy.array
+            Single dimension array with the currents of each wire in NLK.
+        """
+        wires_currents = _np.zeros(len(self._wires))
```
fernandohds564, line 171: this could be simply:
```
return _np.array([wi.current for wi in self._wires])
```

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Returns
+        -------
+        x_space: numpy.array
+            Values of x where the field was computed.
+
+        fieldy: numpy.array
+            Vertical field in [T] at y=0.
+        """
+        x_space = _np.linspace(-12, 12)[None, :]*1e-3
```
fernandohds564, line 220: `x_space` is not a good name here. Maybe `x_pos`?

you could also write:
```
x_pos = _np.linspace(-12, 12)*1e-3
r = _np.hstack([x, 0*x])
```

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+            If input r hasn't shape of type (2 X N).
+        """
+        if r.shape[0] != 2:
+            raise ValueError(
+                "The input position vector r must have shape of type (2, N)")
+        r = r.reshape(2, -1)
+        mu0 = vacuum_permeability
+        r_w = self.position[:, None]  # Wire positions
+        rc = r - r_w  # Cursive r
+        theta = _np.arctan2(rc[1, :], rc[0, :])
```
fernandohds564, line 52: you don't need the `:` in `rc[0, :]`, just do `rc[0]` and `rc[1]`

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+        theta_vec = _np.array([-_np.sin(theta), _np.cos(theta)])
+        rc_norm = _np.linalg.norm(rc, axis=0)[None, :]
+        mag_field = \
+            mu0 * self.current/(2*_np.pi*rc_norm)*theta_vec + r_w
```
fernandohds564, line 56: remove the addition of the radius `r_w`.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+        return mag_field
+
+    @property
+    def position(self):
+        """."""
+        return self._pos
```
fernandohds564, line 62: return a copy here: `self._pos.copy()`

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+from numpy.polynomial.polynomial import polyfit as _polyfit
+import matplotlib.pyplot as _plt
+from mathphys.constants import vacuum_permeability
+from mathphys.beam_optics import beam_rigidity
+
+
+class Wire:
+    """Simulates the magnetic field of a infinite wire.
+    """
+    def __init__(self, position, current):
```
fernandohds564, line 11: add default values for `position=(0, 0)` and `current=1`.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+
+    @current.setter
+    def current(self, current):
+        """."""
+        self._curr = current
+
+    @staticmethod
+    def _check_r_size(r):
+        """."""
+        if r.size == 2:
```
fernandohds564, line 83: this function should also accept a list or a tuple as input. You can do this by adding a line
```
r = _np.asarray(r)
```
before the `if` statement.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+            for i in range(0, wire_positions.shape[0], 2):
+                j = int(i/2)
+                wire_positions[i] = \
+                    _np.array([s1[j]*7.0,  s2[j]*5.21])[None, :]*1e-3
+                wire_positions[i+1] = \
+                    _np.array([s1[j]*10., s2[j]*5.85])[None, :]*1e-3
+
+        # Creating wires
+        self._wires = []
+        for i in range(8):
```
fernandohds564, line 119: you could simply make:
```
currents = _np.zeros(8)
currents[::2] = -current
currents[1::2] = current

self._wires = [Wire() for i in range(8)]
self.positions = wire_positions
self.currents = currents
```

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+    @property
+    def positions(self):
+        """Position coordinates of the wires.
+
+        Returns
+        -------
+        numpy.array
+            Returns a numpy.array of shape (2, 8) with the wires coordinates.
+        """
+        wires_positions = _np.zeros([2, 8])
```
fernandohds564, line 152: this getter could be written as:
```
return _np.array([wi.position for wi in self._wires]).T
```

fernandohds564 (changes requested):

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        if r.shape[0] != 2:
+            raise ValueError(
+                "The input position vector r must have shape of type (2, N)")
+        mag_field = _np.zeros(r.shape)
+        for wire in self.wires:
+            mag_field += wire.calc_magnetic_field(r)
+        return mag_field
+
+    def get_magnetic_field_on_axis(self):
```
murilobalves, line 209: why not have a more generic method `get_vertical_magnetic_field` with `y` input defining the plane, with default value to be 0? And also have `get_horizontal_magnetic_field` with `x` input also defining a plane.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+
+        polynom_b : numpy array of shape (fit_monomials,)
+            Polynom B that describes the NLK kick.
+        """
+
+        nlk_length = 0.45
+        if fit_monomials is None:
+            fit_monomials = _np.arange(10, dtype=int)
+        if strength is None:
+            strength = 1
```
murilobalves, line 267: you do not need to do this if you make `strength=1` in the method arguments.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        x_space = _np.linspace(-12, 12)[None, :]*1e-3
+        y_space = _np.zeros(x_space.shape)
+        r = _np.concatenate([x_space, y_space], axis=0)
+        fieldy = self.calc_magnetic_field(r)[1, :]
+        return x_space[0], fieldy
+
+    @staticmethod
+    def si_nlk_kick(
+            strength=None, fit_monomials=None, plot_flag=False,
```
murilobalves, line 228: I think that the name `strength`might be confusing for this variable... maybe `factor`? it is a unitless multiplicative value. I would expect that `strength` would be the NLK field strength in accelerator units, a kick in mrad for example.

murilobalves (changes requested):

generic_scripts/nlk.py
fernandohds564: `factor` or `scaling` or `scale`, or something similar are indeed better names.

generic_scripts/nlk.py
jquentino: It's a good idea, I will implement it.

generic_scripts/nlk.py
jquentino: Done, thanks!

generic_scripts/nlk.py
```diff
@@ -0,0 +1,304 @@
...
+
+        Returns
+        -------
+        x_pos: numpy.array
+            Values of x where the field was computed.
+
+        fieldy: numpy.array
+            Vertical field in [T] at the plane y=y0.
+        """
+        x_pos = _np.linspace(-12, 12)[None, :]*1e-3
```
fernandohds564, line 214: I think you can simplify this to:
```
x_pos = _np.linspace(-12, 12)*1e-3
y_pos = _np.ones(x_pos.shape) * y0
r = _np.vstack([x_pos, y_pos])
fieldy = self.calc_magnetic_field(r)[1]
return x_pos, fieldy
```

generic_scripts/nlk.py
```diff
@@ -0,0 +1,304 @@
...
+
+        Returns
+        -------
+        y_pos: numpy.array
+            Values of y where the field was computed.
+
+        fieldy: numpy.array
+            Horizontal field in [T] at the plane x=x0.
+        """
+        y_pos = _np.linspace(-12, 12)[None, :]*1e-3
```
fernandohds564, line 236: same here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,304 @@
...
+
+        nlk = NLK()
+        x, mag_field = nlk.get_vertical_magnetic_field()
+
+        brho, *_ = beam_rigidity(energy=3)  # energy in [GeV]
+        integ_field = scale * mag_field * nlk_length  # [T.m]
+        kickx = integ_field / brho  # [rad]
+
+        coeffs = _polyfit(x=x-r0, y=kickx, deg=fit_monomials)
+        fit_kickx = (x[:, None]**fit_monomials[None, :]) @ coeffs
```
fernandohds564, line 291: there is a method in `numpy.polynomial.polynomial` called `polyval`. I think you could used it here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,304 @@
...
+            _plt.figure()
+            _plt.scatter(1e3*(x-r0), 1e3*kickx, label="data points")
+            _plt.plot(
+                1e3*(x-r0), 1e3*fit_kickx, c=[0, 0.6, 0], label="fitted curve")
+            _plt.xlabel('X [mm]')
+            _plt.ylabel("Kick @ 3 GeV [mrad]")
+            _plt.title("NLK Profile")
+            _plt.legend()
+
+        polynom_b = _np.zeros(1 + _np.max(fit_monomials))
```
fernandohds564, line 302: `fit_monomials.max()` would be more pythonic, since it is object oriented.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,290 @@
...
+        current : float, optional.
+            Current of the wires in Amperes, by default 1850 A.
+        """
+        if positions is not None:
+            wire_positions = positions
+        else:
+            wire_positions = _np.zeros([8, 2])
+            s1 = _np.array([1, -1, 1, -1])
+            s2 = _np.array([1, 1, -1, -1])
+            for i in range(0, wire_positions.shape[0], 2):
```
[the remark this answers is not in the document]
jquentino: Done, thanks!

generic_scripts/nlk.py
jquentino: Changed to x_pos.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,304 @@
...
+            raise ValueError(
+                "The input position vector r must have shape of type (2, N)")
+        r = r.reshape(2, -1)
+        mu0 = vacuum_permeability
+        r_w = self.position[:, None]  # Wire positions
+        rc = r - r_w  # Cursive r
+        theta = _np.arctan2(rc[1], rc[0])
+        theta_vec = _np.array([-_np.sin(theta), _np.cos(theta)])
+        rc_norm = _np.linalg.norm(rc, axis=0)[None, :]
+        mag_field = \
```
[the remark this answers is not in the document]
jquentino: No, it doesn't. I will change it.

generic_scripts/nlk.py
jquentino: Yes, in this way became more simple. These lines are a residue of when I was mistakenly trying to use column vectors everywhere.

generic_scripts/nlk.py
jquentino: I read the documentation and apparently `polyval` doesn't have the option to pass which monomials the powers of x were computed. I think this could introduce a bug in the case of `fit_monomials` isn't a consecutive sequence of numbers, i.e. `fit_monomials = [2,4,5,7,9]` or something like that.

fernandohds564: @jquentino take a look at this example. It works fine:
```
import numpy as np
import matplotlib.pyplot as mplt

monoms = np.array([2, 4, 5, 7, 9])
x = np.linspace(0, 2, 1001)

xp = x[::5]
yp = (xp[None, :]**monomials[:, None]).sum(axis=0)

coeffs = np.polynomial.polynomial.polyfit(xp, yp, deg=monoms)
yfit = np.polynomial.polynomial.polyval(x, coeffs)

print(coeffs)
print(coeffs[monoms])
mplt.plot(xp, yp, 'o')
mplt.plot(x, yfit)
mplt.show()
```

the point is that even though you define certain monomials to be fitted by the `polyfit` method, it always returns the coefficients for all the monomials up to the maximum you defined in the input, but it sets them to zero.
This way it is natural to use `polyval` to reconstruct the fitted curve afterwards.

With this in mind, it surprises me that your code runs without errors.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,305 @@
...
+            wire_positions[:, ::2] = _np.array([s1*7, s2*5.21])
+            wire_positions[:, 1::2] = _np.array([s1*10, s2*5.85])
+            wire_positions *= 1e-3
+
+        # Creating wires
+        currents = _np.zeros(8)
+        currents[::2] = -current
+        currents[1::2] = current
+
+        self._wires = [
```
fernandohds564, line 122: I think you don't need this line break here.

generic_scripts/nlk.py
```diff
@@ -0,0 +1,305 @@
...
+        y_pos = _np.linspace(-12, 12)*1e-3
+        x_pos = _np.ones(y_pos.shape) * x0
+        r = _np.vstack([x_pos, y_pos])
+        fieldy = self.calc_magnetic_field(r)[0]
+        return y_pos, fieldy
+
+    @staticmethod
+    def si_nlk_kick(
+            scale=1, fit_monomials=None, plot_flag=False,
+            r0=0.0):
```
fernandohds564, line 244: I think you don't need this line break here.

generic_scripts/nlk.py
jquentino: Solved, thanks!

generic_scripts/nlk.py
jquentino: Yes, you're right. I didn't pay attention to the fact that polyfit returns the coeffs with length equal to the maximum monomial +1, and with zeros in the monomials that weren't set. Thanks for the correction, I made the changes in the last commit.

vellosok75:
Very nice, Jucélio! Maybe plotting the B field vectors with the same length and using a color scheme to encode their magnitudes would make the last plot even nicer!

generic_scripts/nlk.py
```diff
@@ -0,0 +1,303 @@
...
+            _plt.scatter(1e3*(x-r0), 1e3*kickx, label="data points")
+            _plt.plot(
+                1e3*(x-r0), 1e3*fit_kickx, c=[0, 0.6, 0], label="fitted curve")
+            _plt.xlabel('X [mm]')
+            _plt.ylabel("Kick @ 3 GeV [mrad]")
+            _plt.title("NLK Profile")
+            _plt.legend()
+
+        polynom_b = _np.zeros(1 + fit_monomials.max())
+        polynom_b[fit_monomials] = -coeffs/nlk_length
```
fernandohds564, line 302: I think this is a bug...
You need to do:
```
polynom_b[fit_monomials] = -coeffs[fit_monomials]/nlk_length
```
don't you?
Again, it is weird that this didn't raised an error in your tests... Am I missing something?

fernandohds564 (changes requested):

generic_scripts/nlk.py
fernandohds564: nice @jquentino !

murilobalves (approved):

jquentino:
> Very nice, Jucélio! Maybe plotting the B field vectors with the same length and using a color scheme to encode their magnitudes would make the last plot even nicer!

Something like that?

wire_field
nlk_field_log
nlk_field

generic_scripts/nlk.py
jquentino: Yes, its is really a bug. Solved, thanks!

fernandohds564 (approved):
very nice PR @jquentino !! Congrats! 

Please update the code in the description and the figures too.

PR #179, opened 2022-03-11
Orbit Stability Acquision and Analysis

murilobalves:
First iteration in the process of unifying methods implemented in several jupyter notebooks during machine studies regarding the orbit stability acquisition and analysis.

Three classes were implemented in `commisslib.measure_orbit_stability`: 
- `OrbitAnalysis`: with methods to perform filtering, spectral analysis, SVD and plot results. Can be used independently.
- `OrbitAcquisitionParams`: definition of parameters to be used in the acquisition
- `OrbitAcquisition`: class structure to perform the acquisition in the machine, prepare the timing system and configure the BPMs to trigger the acquisition for a given rate. The analysis of data also can be performed in this class, since it is derived from `OrbitAnalysis`. 

Currently there are two types of analysis: orbit and energy. 

The orbit analysis is basically filtering the data around a certain frequency, calculating the spectrum, integrated PSD for each BPM, calculate spacial modes,

The energy analysis is a manipulation of orbit measurements in all BPMs to extract the beam energy deviation as a function of time (allowing to do spectral analysis) and also a non-invasive method for measuring the dispersion function. The following algorithm is used:

1) Filter orbit array around synchrotron frequency with some frequency window to obtain a filtered orbit array.
2) Apply SVD in the filtered orbit array.
3) Find the singular mode whose spacial signature has maximum correlation with a reference dispersion function (measured or model). Let this mode be the dispersion spacial mode.
4) The measured dispersion function can also be obtained from least-squares minimization of the difference between the reference dispersion function and the dispersion spacial mode.
5) Calculate energy deviation over time by projecting the unfiltered orbit data in the direction of dispersion spacial mode.

An example of application of these classes can be found in the script `generic_scripts.energy_orbit_stability_analysis.py`.
To run this script, the user must define the files path with `prefix` variable. For example, in the control room computers, using `prefix=/home/sirius/shared/screens-iocs/data_by_day/` goes to the right path to access the folders `2021-10-25-SI_orbit_stability_characterization/` and `2021-10-25-SI_orbit_stability_characterization/` that contain pickle files with orbit data.

Once the prefix and folders are adjusted properly to access the data, running this script generates these beautiful plots:

image

image

image

image


generic_scripts/energy_orbit_stability_analysis.py
```diff
@@ -0,0 +1,118 @@
...
+        obj.plot_orbit_spectrum(
+            bpmidx=bpmidx, label=lab, fig=fig, axs=axs, color=colors[idx+1])
+    axs[0].legend(
+        loc='upper right', bbox_to_anchor=(1.25, 1.02), prop={'size': 14})
+    axs[0].set_xlim([55, 65])
+    axs[1].set_xlim([55, 65])
+    fig.savefig('spectrum_orbit.png', dpi=300)
+    plt.show()
+
+    title = 'SI-01M2:DI-BPM - Space Modes'
```
fernandohds564, line 106: I didn't understand this title. The space modes are calculated based on data from all BPMs, aren't they?

fernandohds564 (approved):

jquentino (approved):

generic_scripts/energy_orbit_stability_analysis.py
murilobalves: Sure, this title is wrong. Residues from copy+paste... The space mode are obtained from all BPMs.

anacso17 (approved):

RafaelLyra8 (approved):

LuanaVilela (approved):

xresende:
`running this script generates these beautiful plots`. bold! :sunglasses:

xresende (approved):

PR #188, opened 2022-06-24
Vertical dispersion fitting

murilobalves:
Script to fit the vertical dispersion with chromatic skew quadrupoles. 

This script was used to obtain the strengths of normal and skew quadrupoles of  "pymodels.fitted_models". According to its docstring:

"This model was built via fitting of the measured vertical dispersion function with the chromatic skew quadrupoles of the machine. The measurement used for fitting was the matrix measurement of folder: `shared/screens-iocs/data_by_day/2022-05-24-SI_LOCO/` with name: `respmat_endofmay22_bpms_03sector_switching_issue.pickle`. The transverse tunes were also adjusted to match the measured ones."

There are 60 chromatic QS. Thus, the "etay/KsL" response matrix has 60 singular values. A scan of the number of singular values used in the fitting was performed, resulting in:

image

Note that after 41 singular values, the fitting residue is stable while the KsL strengths and beta-beating keep increasing. Therefore it does not make much sense to include more singular values than 41 to improve the fitting. Note that there is a peak in the emittance ratio calculated with the beam envelope formalism at this singular value and also in the minimum tune separation. I still do not understand the reason for this. The emittance ratio calculated with the radiation integrals (which is not adequate for coupled lattices and it is mainly due to the vertical emittance created by the vertical dispersion, not from betatron coupling) evolves smoothly as the singular value number increases. Also, for the singular values matrix, there is a step at SV 40 as can be seen in the plot below: 

image

Note that the minimum tune separation with 41 SVs is about 1%, the same value that we measured. This model reproduces the measured vertical dispersion and the measured global betatron coupling. 

This was the process to define 41 as the number of singular values to be used in the fitting. Initially, fitting the measured vertical dispersion and the global coupling was also the motivation to use 35 SVs, but at the time I did not notice the strange behavior around the SV 35.

The fitted dispersion as compared to the measured one is shown below:
image

The non-negligible difference between the QS strengths obtained with 35 or 41 is shown bellow:

image

The QN and QS strengths can be printed to then be plugged in "pymodels.fitted_models" with the function `print_strengths_fitted_model`. Related to the PR: https://github.com/lnls-fac/pymodels/pull/85

Maybe we should update the QS strengths in the current fitted mode, what do you think @fernandohds564 ?

For the question raised by @xresende related to comparison between ORM of this model and measured. I calculated the ORM of the fitted model with 41 singular values and compared with the measured ORM. The correlation between off-diagonal blocks of measured and fitted matrices is plotted bellow for CHs (blue) and CVs (orange) signatures. 

image

Taking a look on CH and CVs with maximum correlation:
image

Now correctors signatures with minimum correlation:
image

Remember that in this fit **only chromatic skew quadrupoles** were used. In principle, one can use the achromatic QS (the remaining 40 magnets) to match the coupling blocks of ORM and/or control the global betatron coupling to the measured value.


fernandohds564 (commented):
very nice @murilobalves.

xresende:
thanks, @murilobalves . At singular value 35 there is a spike in emittance ratio. Was it just a coincidence that there is no gain in including corrector modes above this singular value? if it is a coincidence, did it not bother you to pick 35 modes, as for they leads to a higher emittance ratio compared to other cases  ?

fernandohds564 (approved):
Yes @murilobalves, I think it is a good idea to make a PR changing these values. However I think it is that PR must include a comparison of the effect of these changes on the longitudinal dynamics of a single-bunch including IBS and wakes.

anacso17 (approved):

xresende (approved):

PR #195, opened 2022-10-21
Rcds

fernandohds564:


vellosok75 (approved):

anacso17 (approved):

apsuite/optimization/rcds.py
```diff
@@ -0,0 +1,457 @@
+"""Robust Conjugate Direction Search Algorithm for Minimization."""
+from threading import Thread as _Thread, Event as _Event
+import logging as _log
+
+import numpy as _np
+import matplotlib.pyplot as _mplt
```
xresende: import not being used

apsuite/optimization/scanning.py
```diff
@@ -8,68 +8,19 @@
...
-
-    def calc_obj_fun(self):
-        """Return arrays with dimension of search space."""
-        raise NotImplementedError
 
     def set_limits(self, upper=None, lower=None):
         """."""
         self._upper_limits = upper
         self._lower_limits = lower
         self.ndim = len(upper)
```
xresende, line 22: I think this should be an explicit property attached to private ttribute `self._dim`. The class interface is cleaner this way (no need to look specific implementation code to access all class interface)

fernandohds564: Thanks! This class is broken, I'm fixing it. This parameter is deprecated with the current standard.

fernandohds564: @xresende, could you take a look again, please.

murilobalves (approved):

PR #225, opened 2023-05-23
DYNAP. ENH: fix percentage sign in plot label and return all the axes on `ph_space.make_figure()`

vellosok75:
I noticed the label for the energy deviation in the plot produced by the `ph_space.make_figure()` method is always broken. This method returns the figure object and some of the axes plotted in it (`phx` and `phy`). I found it useful to return the other axes as well so I could manipulate them (change labels, colors, style or reuse the subplots elsewhere).

### DIFF
--- a/apsuite/dynap/phase_space.py
+++ b/apsuite/dynap/phase_space.py
@@ -264,7 +264,7 @@ class PhaseSpace(_BaseClass):
 
         axx.set_xlabel('X [mm]')
         ayy.set_xlabel('Y [mm]')
-        ade.set_xlabel(r'$\delta$ [%]')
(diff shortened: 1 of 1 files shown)


xresende (approved):

Gabrielrezende-asc (approved):

fernandohds564 (commented):
nice addition! 

I'm thinking here...
wouldn't it be better to return the axes in a list of lists so that it maps the position of the axes in the figure?
Like this:
```python3
return fig, [[phx, phy, tune], [axx, ayy, ade]]
```

what do you think?

vellosok75:
@fernandohds564 Yes, that's great! Done!

fernandohds564 (approved):

PR #231, opened 2023-08-07
Improve switching filtering in OrbitAnalysis

murilobalves:
Required for acquisitions in TbT rate that have switching harmonics

Now scipy.fft is used instead of numpy.fft since it is slightly faster (~20% in my PC. I tested in the control rooms PCs and the performance is the same for both packages).

Example considering  the typical sizes of orbit data executed in my PC:

```
import numpy as np
import scipy as sp

Nsamps = 100_000
Nbpms = 160
mat = np.random.rand(Nsamps, Nbpms)

%timeit np_fft = np.fft.rfft(mat)
# result:  56.1 ms ± 107 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
%timeit sp_fft = sp.fft.rfft(mat)
# result: 44.5 ms ± 110 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

### DIFF
--- a/apsuite/commisslib/measure_orbit_stability.py
+++ b/apsuite/commisslib/measure_orbit_stability.py
@@ -38,6 +38,7 @@ class OrbitAnalysis:
         self._sampling_freq = None
         self._switching_freq = None
         self._rf_freq = None
+        self._frequency_band_filter = None
         self.analysis = dict()
         self.orm_client = _sconf.ConfigDBClient(config_type='si_orbcorr_respm')
         if self.fname:
@@ -126,6 +127,15 @@ class OrbitAnalysis:
         self._switching_freq = val
 
     @property
+    def frequency_band_filter(self):
+        """."""
+        return self._frequency_band_filter
+
+    @frequency_band_filter.setter
+    def frequency_band_filter(self, val):
+        self._frequency_band_filter = val
+
+    @property
     def rf_freq(self):
         """."""
         return self._rf_freq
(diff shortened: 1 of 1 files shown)


anacso17 (approved):

xresende (approved):

apsuite/commisslib/measure_orbit_stability.py
```diff
@@ -534,15 +551,64 @@ def _get_switching_freq(self):
 
     # static methods
     @staticmethod
-    def filter_matrix(matrix, fmin=0, fmax=None, fs=1):
-        """."""
-        if fmax is None:
-            fmax = fs/2
+    def filter_matrix(matrix, fmin, fmax, fsampling, keep_within_range=True):
```
fernandohds564, line 554: I don't like this name. Data does not need to be a matrix to be filtered. It filters each vector individually, so it is intrinsically a 1D algorithm. The only difference is that it is vectorized.

apsuite/commisslib/measure_orbit_stability.py
```diff
@@ -534,15 +551,64 @@ def _get_switching_freq(self):
...
+            freq_sampling (float): Sampling frequency of the input signal.
+            freq_switching (float): Switching frequency to be filtered out.
+
+        Returns:
+            numpy.ndarray: Signal with the switching frequency removed, same
+            shape as the input.
+
+        """
+        # Calculate the number of samples per switching cycle
+        sw_sample_size = round(freq_sampling/freq_switching)
```
fernandohds564, line 595: why using the frequencies here for this approximated calculation? why not using the integer divisors that the BPM IOC provides?

apsuite/commisslib/measure_orbit_stability.py
```diff
@@ -190,17 +200,20 @@ def calc_integrated_spectrum(self, spec, inverse=False):
...
-        orby = orby or self.orby.copy()
         fs = self.sampling_freq
         fsw = self.switching_freq
-        fil_orbx, freq = self.filter_matrix(
-            orbx, fmin=0, fmax=fsw*0.9, fs=fs)
-        fil_orby, _ = self.filter_matrix(
-            orby, fmin=0, fmax=fsw*0.9, fs=fs)
-        return fil_orbx, fil_orby, freq
+        frev = self.rf_freq/self.HARM_NUM
+        if fs == frev:
```
fernandohds564, line 208: this is not a good way to identify TbT data. Try to do this avoiding this float comparison.
I also think this filtering algorithm is not particular to TbT data. It could also be used for ADC or FOFB rates.

apsuite/commisslib/measure_orbit_stability.py
```diff
@@ -125,6 +126,15 @@ def switching_freq(self):
     def switching_freq(self, val):
         self._switching_freq = val
 
+    @property
+    def frequency_band_filter(self):
+        """."""
+        return self._frequency_band_filter
+
+    @frequency_band_filter.setter
+    def frequency_band_filter(self, val):
```
fernandohds564, line 135: this name is not intuitive. This is not a filter, it is just bandwidth.

fernandohds564 (changes requested):

apsuite/commisslib/measure_orbit_stability.py
murilobalves: This calculation should be exact, actually. The acquisition and sampling rates are not being recorded directly in the data acquisition. However, the sampling and switching frequencies are recorded with the FamBPMs's methods `get_sampling_frequency` and `get_switching_frequency`, which takes the RF frequency (also recorded in the data) at the moment of acquisition  to calculate the corresponding frequencies, using the aforementioned integer divisors from the BPM IOC. In this way, the calculation in the code should be equivalent to using the integer divisors directly.

apsuite/commisslib/measure_orbit_stability.py
murilobalves: You're right. Following what we discussed, now it will filter switching only if switching mode was on during acquisition and the sampling frequency is greater than switching frequency (FOFB, ADC, TbT rates).

apsuite/commisslib/measure_orbit_stability.py
murilobalves: With the new switching filtering method this attribute is no longer necessary and I deleted it.

apsuite/commisslib/measure_orbit_stability.py
murilobalves: Right. I renamed it to `filter_orbit_frequencies`

fernandohds564 (approved):

Gabrielrezende-asc (approved):

PR #234, opened 2023-08-10
Make it possible to acquire sum data in OrbitAcquisition

anacso17:


fernandohds564 (approved):
Nice @anacso17 !
It would be nice to also have a way of getting the antennas data. Maybe we could generalize FamBPMs and this class for this purpose.

Gabrielrezende-asc (approved):

murilobalves (approved):

PR #236, opened 2023-08-15
Print BPM name and index when orbit acquisition failed

murilobalves:


anacso17:
Maybe we could change FamBPMs to return name of the problematic BPMs ?

fernandohds564 (approved):

anacso17:
> > Maybe we could change FamBPMs to return name of the problematic BPMs ?
> 
> The first thought was including this in the FamBPMs but then I preferred to locate the change in this apsuite class and keep the FamBPMs interface unchanged. It is very simple to get the name of the problematic BPM once you already have a FamBPMs object. What do you think @anacso17 @fernandohds564 ?

I see, this would be a big change in FamBPMs interface... We can discuss offline in other opportunity (:

anacso17 (approved):

PR #245, opened 2023-10-15
meas_fofb_sysid.py: changes in get_levels_bpms_from_svd

guilhermerc:
As an attempt to prevent correctors from saturating, PRBS amplitudes are being scaled based on the mode's singular value when computing the levels for BPMs' excitation. Also, a configurable lower bound amplitude can be setted so the excitation still produces reasonable correctors' strengths.

### DIFF
--- a/apsuite/commisslib/meas_fofb_sysid.py
+++ b/apsuite/commisslib/meas_fofb_sysid.py
@@ -247,7 +247,7 @@ class FOFBSysIdAcq(_BaseClass):
         lvls1 = off + amp * vs
         return lvls0[:-1], lvls1[:-1]
 
-    def get_levels_bpms_from_svd(self, lvl0, lvl1):
+    def get_levels_bpms_from_svd(self, lvl0, lvl1, ampmin=0):
         """Get levels from SVD for BPMs devices.
 
         Args:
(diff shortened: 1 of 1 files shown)


apsuite/commisslib/meas_fofb_sysid.py
```diff
@@ -266,17 +266,24 @@ def get_levels_bpms_from_svd(self, lvl0, lvl1):
...
 
         if self.params.svd_levels_regularize_matrix:
             respm = self._calc_matrix_regularization(respm)
         respm = self._convert_respmat_from_phys2hard_units(respm)
 
-        u, *_ = self._calc_svd(respm)
+        u, s, _ = self._calc_svd(respm)
         us = u[:, singval]
         us /= _np.abs(us).max()
+        ss = s[singval]
```
fernandohds564, line 280: There's a bug here. You need to normalize before selecting, otherwise `ss` will be always one.

fernandohds564 (changes requested):

apsuite/commisslib/meas_fofb_sysid.py
guilhermerc: Thanks for the catch, @fernandohds564. 1bb8002b4c4cb7e78cd6f049024c9a0a780d172f fixes it. Are ok with that?

anacso17 (approved):

fernandohds564 (approved):

PR #267, opened 2024-03-05
COMMISS.FOFB.ENH: use largest amp possible.

ericonr:
If we scale the singular vectors used for closed loop identification starting from the largest amplitude possible for the last mode, we can try and guarantee the best possible SNR for all modes. It then becomes necessary to add an ampmax argument as well.


apsuite/commisslib/meas_fofb_sysid.py
```diff
@@ -278,12 +278,12 @@ def get_levels_bpms_from_svd(self, lvl0, lvl1, ampmin=0):
         us = u[:, singval]
         us /= _np.abs(us).max()
```
ericonr, line 279: I'm wondering if strictly speaking we shouldn't remove this division here... The vectors are normalized, so in order to keep the relationship between them and the control effort, we should keep them as is, instead of scaling using the biggest value, since that's kind of arbitrary (some modes can be more spread out than others, for example).

@fernandohds564 @guilhermerc what do you think? (adding you two since you aren't reviewers)

juliacscarvalho (approved):

apsuite/commisslib/meas_fofb_sysid.py
fernandohds564: I think at the time this was implemented you guys wanted to respect the limits defined by the levels, but I agree that it is better to remove this line to make things easier.

apsuite/commisslib/meas_fofb_sysid.py
```diff
@@ -276,14 +277,13 @@ def get_levels_bpms_from_svd(self, lvl0, lvl1, ampmin=0):
 
         u, s, _ = self._calc_svd(respm)
         us = u[:, singval]
-        us /= _np.abs(us).max()
         ss = s[singval]
-        ss /= _np.abs(s).max()
+        ss /= _np.abs(s).min()
```
fernandohds564, line 281: isn't it too dangerous to divide by the minimum here? What if we have an almost degenerate matrix?

I don't know if I got the advantage of this new way of limiting by above, instead of old way, where the distortion was limited by below. The other one seemed more intuitive for me...

ericonr: > isn't it too dangerous to divide by the minimum here? What if we have an almost degenerate matrix?

Shouldn't this be avoided by the minimum singular value setting? It seems to be implemented in `_calc_matrix_regularization`.

> I don't know if I got the advantage of this new way of limiting by above, instead of old way, where the distortion was limited by below. The other one seemed more intuitive for me...

The issue is that the magnitude of the singular values covers a very wide range. With the original implementation, in order to get reasonable SNR for the lowest SVs, we'd have to configure way too high distortion profiles for the highest SVs. With this implementation, we configure a known good value which shouldn't saturate voltage for the lowest mode, and scale it up to get better SNR for the SVs which allow it, until a max limit where increasing the distortion doesn't make a huge difference.

(I can add this explanation to the commit, if you think it helped)

fernandohds564: Ok, if this code sees a regularized matrix, the chances of having problems are smaller.

If you guys think it is more intuitive this way, ok then.  
There's no need to edit the commit.

fernandohds564 (approved):

guilhermerc (approved):

juliacscarvalho (approved):

anacso17 (approved):

PR #268, opened 2024-03-21
Temperature Excursion

fernandohds564:
Add script in folder `generic_scripts` to calculate temperature excursion of electron beam collision with vacuum chamber.

### DIFF
--- a/generic_scripts/vchamber_temp_excursion_induced_by_electrons.py
+++ b/generic_scripts/vchamber_temp_excursion_induced_by_electrons.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""Temperature excursion induced by electron beam.
+
+This script calculates the temperature excursions of SIRIUS vacuum chambers
+induced by the collision with the eletron beam.
+
+The dose generated is calculated based on equations from [1] and the
+temperature excursion is calculated based on [2].
+
+The materials properties are taken from [3].
+
+The beam equilibrium parameters for the storage ring model with vertical
+dispersion and coupling fitted to measurements are used.
+
+[1] Jeffrey Dooling, private communication.
+[2] equation 2.4 of APS-U Final Design Report. Available at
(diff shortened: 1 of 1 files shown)


murilobalves (approved):

Gabrielrezende-asc (approved):

vellosok75 (approved):

otavio-silveira (approved):

generic_scripts/vchamber_temp_excursion_induced_by_electrons.py
```diff
@@ -0,0 +1,192 @@
...
+        f_rev = f_rf / harm  # [Hz]
+        charge = self.current / f_rev  # [C]
+
+        self.eqpar = pa.optics.EqParamsFromBeamEnvelope(mod)
+        self.spos = np.array(pa.lattice.find_spos(mod, indices='closed'))
+        sigx = self.eqpar.sigma_rx
+        sigy = self.eqpar.sigma_ry
+
+        # energy density in [J/m^2]
+        self.energy_density = charge * energy / 2 / np.pi / sigx / sigy
```
xresende, line 138: maybe the class interface should make explicit that various calculated properties are for normal beam collision, and not grazing collisions.

xresende (approved):

generic_scripts/vchamber_temp_excursion_induced_by_electrons.py
fernandohds564: I'm not sure if that is True. The particles will penetrate the chamber and the cross-section area of the cylinder formed along the depth of the material is the same for both types of collision. Since this is an open question, I prefer not to add any comment.

VitorSouzaLNLS (approved):

generic_scripts/vchamber_temp_excursion_induced_by_electrons.py
fernandohds564: we talked offline and agreed not to add any comment.

PR #269, opened 2024-04-12
TbT data measurement & analysis classes 

vellosok75:
WORK IN PROGRESS: module for measuring and analyzing linear optics from turn-by-turn BPM acquisitions.

The measurement script has been tested online and works well. Only linear optics analysis is available for now (beta beating and phase advance errors), as probed by either sinusoidal fitting, PCA or ICA. Equilibrium parameters fitting (energy spread and emittances) is not yet implemented.


apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,209 @@
...
+            of Ref. [1]"""
+        action = _np.sum(amplitudes**4)
+        action /= _np.sum(amplitudes**2 * nominal_beta)
+        beta = amplitudes**2/action
+        return beta, action
+
+class TbTDataParams(_AcqBPMsSignalsParams):
+    """."""
+
+    def __init__(self, hkick=None, vkick=None):
```
murilobalves, line 133: Allowing inputs is not common in our Params classes. Additionally, this is a potential bug. If the user don't set this optional argument, hkick and vkick will be None and then `prepare_pingers` will set None to the pingers strength, which I think will lead to an error. 

A possible option is that you can initialize `self._hkick = None` and at some point check it is still None, do nothing with the pingers.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,209 @@
...
+    def fname(self, val):
+        self._fname = val
+
+    def prepare_pingers(self):
+        """."""
+        pingh, pingv = self.devices['pingh'], self.devices['pingv']
+        params = self.params
+        hkick, vkick = params['hkick'], params['vkick']
+        pingh.strength = hkick / 1e3   # [urad]
+        pingv.strength = vkick / 1e3   # [urad]
```
murilobalves, line 206: It is good to keep our measurement classes as transparent as possible: before the measurement, store the values of parameters that will be changed, change it for the measurement, and then restore the initial state.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,217 @@
...
+        hkick, vkick = self.params['hkick'], self.params['vkick']
+        pingh.strength = hkick / 1e3   # [urad]
+        pingv.strength = vkick / 1e3   # [urad]
+        # MISSING: set to listen to the same event as BPMs
+
+    def do_measurement(self):
+        currinfo = self.devices['currinfo']
+        self.prepare_pingers()
+        current_before = currinfo.current()
+        self.acquire_data()
```
murilobalves, line 213: Maybe just leave a comment briefly describing important data that is stored here from `AcqBPMsSignals`. I would suggest to measure the tunes but then I saw that is already measured in the parent class.

murilobalves (changes requested):

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,217 @@
...
+import numpy as _np
+from scipy.optimize import curve_fit as _curve_fit
+import matplotlib.pyplot as _mplt
+import datetime as _datetime
+from siriuspy.devices import PowerSupplyPU
+
+from .meas_bpms_signals import AcqBPMsSignals as _AcqBPMsSignals, \
+    AcqBPMsSignalsParams as _AcqBPMsSignalsParams
+
+class TbTDataAnalysis(_AcqBPMsSignals):
```
murilobalves, line 11: We typically write the classes following the workflow: Params -> Measurement -> Analysis. It is just a matter of organization, it doesn't affect how the classes work.

apsuite/commisslib/measure_tbt_data.py
vellosok75: Thanks! Hadn't thought about this possibility. Removed these properties as inputs and added checkings to avoid the  None type to be set as the magnets strengths.

apsuite/commisslib/measure_tbt_data.py
vellosok75: Thanks for the feedback, Murilo. I hadn't yet checked out the acquisition script you mentioned friday. I did it this morning and brought some patterns from there, as well as working on you other suggestions. Thanks!

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+        self.data['current_before'] = current_before
+        self.data['current_after'] = self.data.pop('stored_current')
+
+    def get_default_fname(self):
+        """."""
+        hkick, vkick = self.params.hkick, self.params.vkick
+        tm = self.data['timestamp']
+        fmt = '%Y-%m-%d-%H-%M-%S'
+        tmstp = _datetime.datetime.fromtimestamp(tm).strftime(fmt)
+        stg = f'hkick_{int(round(hkick)):3d}_vkick_{int(round(vkick)):3d}'
```
fernandohds564, line 171: Add information regarding which kicks were kicked.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+    def fname(self):
+        """."""
+        return self._fname
+
+    @fname.setter
+    def fname(self, val):
+        self._fname = val
+
+    @property
+    def trajx(self):
```
fernandohds564, line 201: you should get all these properties from the data loaded from file. And remember to return a copy of them, to guarantee data integrity.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+    """."""
+
+    def __init__(self, filename='', isonline=False):
+        """Analysis of linear optics using Turn-by-turn data."""
+        self.params = TbTDataParams()
+        self.isonline = isonline
+        self._ispost_mortem = False
+
+        self._fname = filename
+        self._trajx, self._trajy = None, None
```
fernandohds564, line 186: you don't need these properties. See comment below.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+        self._trajsum = None
+
+        # load if fname, load method
+
+    @property
+    def fname(self):
```
fernandohds564, line 192: As we talked offline, I don't think this property is necessary.

fernandohds564 (changes requested):

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+        if hkick is not None:
+            pingh.strength = hkick
+        else:
+            pingh.strength = self.params.hkick / 1e6
+        _time.sleep(0.5)
+        if vkick is not None:
+            pingv.strength = vkick
+        else:
+            pingv.strength = self.params.vkick / 1e6
+        _time.sleep(0.5)
```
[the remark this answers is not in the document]
vellosok75: I discussed this with @murilobalves  and @xresende  yesterday. From what I understood, monitors for for pulsed magnets can vary significantly from the set-point values and using them may not be very reliable. Additionally, the `PowerSupplyPU` class has only voltage monitors currently, not strength monitors. 

Do you have any suggestions? Do you think it's worthwhile using monitors and convert voltages to strengths?

apsuite/commisslib/measure_tbt_data.py
vellosok75: I'll be keeping it for now, following the same patterns of the `OrbitAnalysis` class

apsuite/commisslib/measure_tbt_data.py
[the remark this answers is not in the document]
vellosok75: Sorry, I overlooked the `set_strength` function from `PSDev`. Thanks for bringing it up.

apsuite/commisslib/measure_tbt_data.py
vellosok75: This info will be conveyed by the `"hkick"` or `"vkick"` strings and their corresponding values in the default fname. If only one of them is present, then only the corresponding magnet has pulsed. If both are present, both magnets pulsed.

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def get_default_fname(self):
+        """."""
+        hkick, vkick = self.params.hkick, self.params.vkick
+        tm = self.data['timestamp']
+        fmt = '%Y-%m-%d-%H-%M-%S'
+        tmstp = _datetime.datetime.fromtimestamp(tm).strftime(fmt)
+        stg = f'hkick_{int(round(hkick)):3d}_vkick_{int(round(vkick)):3d}'
+        stg += f'_urad_{tmstp}'
+        return stg
```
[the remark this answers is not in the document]
vellosok75: Thought of "kicked_data", but I'm not very happy with it. If you guys have any other suggestions let me know.

apsuite/commisslib/measure_tbt_data.py
fernandohds564: I liked the idea of embedding this info in the `hkick` and `vkick` arguments, but maybe it would be better to always keep both on the name. The kicker that didn't pulse could have a different value, though. I was thinking on something like `none` or `nopulse` or `standby` or `inactive` or something else. From these suggestions, I think I prefer `inactive`, but I think you will come up with something better :smile:

fernandohds564 (changes requested):

apsuite/commisslib/measure_tbt_data.py
```diff
@@ -0,0 +1,314 @@
...
+        )
+
+    def do_measurement(self):
+        """."""
+        currinfo = self.devices["currinfo"]
+        init_timing_state = self.get_timing_state()
+        init_magnets_strength = self.get_magnets_strength()
+        current_before = currinfo.current()
+        self.prepare_timing()
+        self.set_magnets_strength()  # gets strengths from params
```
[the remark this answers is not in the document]
vellosok75: Thanks, Fernando. I made some changes in the measurement flow, magnets settings and waiting times. Could you take a look again and let me know if you have any other suggestions?

apsuite/commisslib/measure_tbt_data.py
xresende: Is it possible for the measurement to be interrupted ? If yes, @vellosok75 did you implement restoration in this situation, apart from restoration at completion ?

fernandohds564:
@vellosok75, what is missing here for merge?

PR #275, opened 2024-07-23
Saving positions & objective function values during each evaluation + other changes

vellosok75:
With these changes:
## Base optimization class
- optimization algorithms automatically save and store the values of the objective function and knobs whenever the objective function is measured.
- the objective function output has its shape determined by the shape of the input position-array and by the nature of the optimization problem at hand (multi- or single-objective). For single-objective, single-individual algorithms, such as RCDS, a scalar value is returned by default.
- the concept of the accumulated optima found during the evaluations is introduced by storing indices of the optima along the evaluations history
- bugs related to the duplicated evaluation of the objective function when `nan`s are present in the positions were fixed.
- a default visualization method for accompanying the history of optimization  was introduced to be re-derived in child classes, suiting the specific needs for different optimization algorithms. Also added was a method for displaying slices of the parameter space for a given knob (see example for RCDS below).
## RCDS
- Introduced properties to keep track of RCDS iterations and the number of evaluations during a given iteration
- fixed a minor issue with breaking the iterations loop, ensuring correct calculation of relevant quantities prior to finalization
- introduced plotting methods for accompanying the optimization history, as shown below:
image
- and a plotting method to display slices of the parameter space during evaluations
image

## Other files
- mostly conforming other optimization the algorithms to the changes in the base class
- I saw no need to propagate these changes to the experiment classes derived from RCDS. It is true that some functionalities are now duplicated (storing objective function values and knobs positions every evaluation), however, I see these things as separate: RCDS will only be aware of the state of the knobs & objective function at each evaluation, while in the experiments we are usually interested in the state of the system at hand as a whole, storing the state in the `data` dict. 

# Snippet to reproduce these plots
Create params and optimization problem classes
```python
class ParaboloidOptimizationParams(RCDSParams):

    def __init__(self):
        super().__init__()`

class ParaboloidOptimization(RCDS):
    """Minimize a noisy paraboloid."""

    def __init__(self, use_thread=False, isonline=False):
        super().__init__(use_thread, isonline)
        self.params = ParaboloidOptimizationParams()


    def objective_function(self, pos):
        return np.dot(pos, pos) + self.params.noise_level * np.random.randn()
```
Instantiate, set params,
```python
opt = ParaboloidOptimization()
# set params
opt.params.limit_lower = [-3, -5]
opt.params.limit_upper = [4, 6]
opt.params.max_number_evals = 500
opt.params.max_number_iters = 5
opt.params.initial_stepsize = 0.0100
opt.params.noise_level = 1e-2
opt.params.tolerance = 1e-5
opt.params.initial_position = np.random.uniform(
    low=opt.params.limit_lower, high=opt.params.limit_upper
)
opt.params.initial_search_directions = np.eye(2)
```
Run optimzation
```python
opt.start()
```
Plot
```python
fig, ax = opt.plot_history()
fig, ax = opt.plot_knobspace_slice()
```


murilobalves (approved):

apsuite/optimization/base.py
```diff
@@ -306,3 +336,21 @@ def _target_func(self):
...
+        """
+        funcs = _np.array(self.objfuncs_evaluated)
+
+        optima = []
+        for i in range(len(funcs)):
+            funcsi = funcs[:i+1]
+            funcsi = funcsi[~_np.isnan(funcsi)]
+            optima.append(funcsi.min())
+
+        self.idcs_cumulated_optimum = _np.isin(funcs, optima).nonzero()[0]
```
fernandohds564, line 356: I think there are two problems here:
 1 - This method is not general enough to handle multi-objective and multi-population algorithms;
 2 - In the case of single-objective functions, I think this algorithm over-complicate the task and it also looks to be very slow. Have you considered the simple approach below:
```suggestion
        optima = []
        mini = _np.inf
        for i, fun in enumerate(funcs):
            if not _np.isnan(fun) and fun < mini:
                optima.append(i)
                mini = fun
        self.idcs_cumulated_optimum = _np.array(optima)
```

apsuite/optimization/base.py
```diff
@@ -211,8 +211,11 @@ def __init__(self, params, use_thread=True, isonline=True):
         self.use_thread = use_thread
 
         self._num_objective_evals = 0
-        self.best_positions = _np.array([], ndmin=2)
-        self.best_objfuncs = _np.array([], ndmin=2)
+        self.positions_evaluated = []
+        self.positions_best = []
+        self.objfuncs_evaluated = []
+        self.objfuncs_best = []
+        self.idcs_cumulated_optimum = None
```
fernandohds564, line 218: Maybe, to keep the standardization of attributes names, this attribute could be called: `objfuncs_cumul_optima_idcs`. What do you think?

apsuite/optimization/rcds.py
```diff
@@ -431,15 +454,186 @@ def _optimize(self):
...
+
+        colorbar = fig.colorbar(scatter, ax=ax)
+        colorbar.set_label('objective function value')
+
+        ax.legend()
+        ax.set_title("parameter space slice")
+
+        return fig, ax
+
+    def get_cumulated_optimum(self):
```
fernandohds564, line 616: why isn't this method at the base class?

apsuite/optimization/rcds.py
```diff
@@ -431,15 +454,186 @@ def _optimize(self):
...
+        ax.set_ylabel("knobs values")
+        ax.set_xlabel("evaluations")
+        ax.legend()
+
+        fig.tight_layout()
+        fig.show()
+
+        return fig, axs
+
+    def plot_knobspace_slice(self, dir_idcs=(1, 2)):
```
fernandohds564, line 546: couldn't this method be defined at the base class?

fernandohds564 (commented):
The PR looks good. Thanks @vellosok75!

I left a few comments and suggestions in the code.

murilobalves (approved):

apsuite/optimization/base.py
vellosok75: Thanks for the suggestion, Fernando. Indeed it was too complicated. I have implemented your suggestion and also generalized it to the search of cumulative optima along a certain objective and certain individual for multi-objective and multi-individual algorithms. I also generalized the `plot_knobspace_slice`.

apsuite/optimization/base.py
vellosok75: Done! Also renamed other methods related to "cumulative optima"

apsuite/optimization/rcds.py
vellosok75: With the generalization, now it is.

apsuite/optimization/rcds.py
vellosok75: Same here. Now it is.

fernandohds564 (approved):

xresende (approved):

PR #283, opened 2024-10-25
add idcs_out argument

Gabrielrezende-asc:
This PR opens the possibility to exclude some specific quadrupoles from tune correction. This can be useful when we symmetrize optics using the closest quadrupoles to an ID and then we want to correct tunes without changing this local simmetrization.


Gabrielrezende-asc:
Some plots to compare the effect of both correction methods:
This is the optics symmetrization after the insertion of IVU18 at SB08:
imagem

This is the tune correction after symmetrization using all quadrupoles:
imagem (1)

This is the tune corrector after symmetrization using all quadrupoles except for the triplet closest to the ID:
imagem (2)

It is observed a reduction on beta beating.

fernandohds564 (approved):

murilobalves (approved):

vellosok75 (approved):

xresende (approved):

PR #284, opened 2024-11-01
Improve AC ORM processing for RF frequency step measurement

vellosok75:
During the last studies, we noticed that the processing  for the RF frequency step method for measuring the RF column of the ORM was broken. 

The previous approach consisted on fitting a square waveform to the acquired orbit to identify the step-transitions in orbit due to step RF freq changes. However, the estimated points of transition were often wrong.

The approach I introduce here relies on identifying the transitions points by finding where the orbit is below/above half the estimated orbit change for a given frequency step.

### DIFF
--- a/apsuite/commisslib/meas_ac_orm.py
+++ b/apsuite/commisslib/meas_ac_orm.py
@@ -340,7 +340,7 @@ class MeasACORM(_ThreadBaseClass):
 
     def save_loco_input_data(
             self, respmat_name: str, overwrite=False, save2servconf=False,
-            extra_kwargs=None):
+            extra_kwargs=None, matrix=None):
         """Save in `.pickle` format a dictionary with all LOCO input data.
 
         Args:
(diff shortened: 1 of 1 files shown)


fernandohds564 (approved):
Very good @vellosok75 ! Did you test this with existing data?

murilobalves (approved):

vellosok75:
> Very good @vellosok75 ! Did you test this with existing data?

I tested this new processing with high-current acquisitions (~100 mA) and it worked well. The RF column was successfully reconstructed and agreed well with the reference resp. mat. at the time, except for the signature in the vertical orbit (more details below). Unfortunately I do not the the results easily accessible right now.
 
What I can show is the more delicate case which motivated the use of the smoothing filter prior to the transitions identification. 

At low currents, the signal-to-noise ratio is low for the RF step method. The figure below, measured at 9 mA, makes this very clear.
avg_orbdist_rfline

In this scenario the transitions identification applied directly to the raw avg. orbit failed because of the noisy readings, but worked fine when using the smoothed signal. 

The figure below shows the reconstructed RF column using this method.
acorm_vs_refrespmat
Again I stress the different signature along the vertical BPMs. I still don't understand it, but if my memory is not fooling me I believe we have seen a similar behavior to this in past AC acquisitions, so I wanted to hear from you guys what do you think. 

The latest ref-respmats have been measured with BPMs acquisitions during DC RF freq. steps. The step measurement of the AC ORM does the same, the difference being that a single BPMs acquisition is fired and records both the positive and negative distortions in orbit, which is then processed with the new processing I mentioned. So it's strange the results are not quite there for the vertical BPMs. I'll try to examine the processing carefully today to check if there's anything wrong...

vellosok75:
Hey guys,

@fernandohds564 reminded me that these two dispersions I'm comparing are not quite comparable for validating the new processing. The ref-respmat (orange in the plot above) is measured after convergence of the LOCO iterations and therefore has a corrected vertical dispersion. However, the measurement I processed and plotted in blue was taken soon after applying the first LOCO corrections, starting from zero trim coils. 

Unfortunately I could not find more raw data of ORM acqusitions to reconstruct the ORMs and validate this new processing, but I could compare the ORM measured at 6 mA soon after turning-off the trims and the ORM acquired at 9 mA soon after applying the 1st round of LOCO corrections. In this case, the vertical dispersions are comparable.

orm_comparison

I ran the new RF column processing step-by-step and it seems to work the way it supposed to be. Therefore, I believe the vertical signature I mentioned in the previous comment is actually just the uncorrected vertical dispersion.

PR #286, opened 2024-11-21
Fix orbcorr cavity

Gabrielrezende-asc:
A possible problem in the orbit correction class:
1) The cavity was always being set as True regardless of the initial state of the model, and the original state was not being restored.
2) The cavity must be set to True when calculating the kicks to correct the orbit if we choose to use the RF correction, so I added a line to set the cavity to True in this function and then restore the original value.

### DIFF
--- a/apsuite/orbcorr/orbit_correction.py
+++ b/apsuite/orbcorr/orbit_correction.py
@@ -71,8 +71,10 @@ class OrbitCorr:
         else:
             raise ValueError('Corretion system must be "SOFB" or "FOFB"')
         if self.params.enblrf and model.cavity_on is False:
-            raise Exception("It is necessary to turn the cavity on if it is"
-                            " to be used in correction..")
+            raise ValueError(
+                "It is necessary to turn on the cavity if it is "
+                "to be used in correction."
+            )
 
         self.params.enbllistch = _np.ones(
             self.respm.ch_idx.size, dtype=bool)


xresende:
should we keep modifying the model under the hood for the use of algorithms to make sense or should we raise an explanatory exception educating the user on the propor use of the algorithm ? 

@murilobalves , @fernandohds564 , @vellosok75 , @Gabrielrezende-asc 

another concern: potential inconsistencies because of temporary modifications of the model when threads that use the model are running.

fernandohds564:
I think we should drop this behavior and raise exceptions when the model state doesn't match what is expected.

xresende:
@Gabrielrezende-asc 

> I think we should drop this behavior and raise exceptions when the model state doesn't match what is expected.

agreed

xresende:
@murilobalves , @vellosok75 , do you want to take a position on this ?

@Gabrielrezende-asc , if others do not object, I think we can proceed with what fernando and I agreed.

xresende (changes requested):
If no one object we should implement what was discussed in the conversation.

vellosok75:
> @murilobalves , @vellosok75 , do you want to take a position on this ?
> 
> @Gabrielrezende-asc , if others do not object, I think we can proceed with what fernando and I agreed.

I agree. I prefer raising exceptions.

apsuite/orbcorr/orbit_correction.py
```diff
@@ -262,6 +261,8 @@ def set_delta_kicks(self, dkicks):
     def set_kicks(self, kicks):
```
vellosok75, line 261: Also, shouldn't we be working with the `set_delta_kicks` method in `correct_orbit`? I believe the `set_kicks` is incorrect since the pseudo-inversion is giving us deltas, right?

murilobalves:
> should we keep modifying the model under the hood for the use of algorithms to make sense or should we raise an explanatory exception educating the user on the propor use of the algorithm ?
> 
> @murilobalves , @fernandohds564 , @vellosok75 , @Gabrielrezende-asc
> 
> another concern: potential inconsistencies because of temporary modifications of the model when threads that use the same model are running.

I agree with raising exceptions.

Gabrielrezende-asc:
Ok guys, as soon as I finish tasks related to the VPU commissioning I will continue this PR

apsuite/orbcorr/orbit_correction.py
```diff
@@ -71,6 +70,9 @@ def __init__(self, model, acc, params=None, corr_system='SOFB'):
             pass
         else:
             raise ValueError('Corretion system must be "SOFB" or "FOFB"')
+        if self.params.enblrf and model.cavity_on is False:
+            raise Exception("It is necessary to turn the cavity on if it is"
+                            " to be used in correction..")
```
fernandohds564, line 75: I think `Exception` is too general. What do you think of the suggestion below?
```suggestion
            raise ValueError(
                "It is necessary to turn on the cavity if it is "
                "to be used in correction."
            )
```

murilobalves (approved):

PR #299, opened 2025-06-09
Add waits & status checkings for methods used when setting LOCO-fitted strengths

vellosok75:
- uses powersupply's `set_strength` method
- prior to these changes, the different `apply` functions returned the initial magnets strengths. With these modifications, they return whether the strengths have been successfully set or timed-out.

Companion of https://github.com/lnls-sirius/scripts/pull/179


fernandohds564:
I left some comments on the other PR, since this branch is already merged there, right?

fernandohds564 (approved):
@vellosok75, in the future, try to separate commits where you format code from commits where you actually change the code. Mixing these things make it very difficult to review the PR.

vellosok75:
@murilobalves I added an additional `ref_strength` arg to the methods so that the we can calculate the strengths to apply with respect to a fixed reference, instead of with respect to the current state of the machine. In this manner, with the fixed reference, we can apply strengths step-by-step by varying the percentage at each application. For example, 20%, then reapply with 40%, 50% etc... If None is passed to this argument, the methods are unchanged.

xresende (approved):

anacso17:
@vellosok75 , same question here about deploying. Please, update version here to `2.56.0`.

PR #301, opened 2025-07-18
Save `bpms2dobba` in BBA data file

vellosok75:
Partner of https://github.com/lnls-sirius/scripts/pull/181

Saves `bpms2dobba` in BBA file so that progress from previous measurements can be recovered by comparing measured vs to-be measured BPMs.

### DIFF
--- a/apsuite/commisslib/measure_bba.py
+++ b/apsuite/commisslib/measure_bba.py
@@ -1555,6 +1555,7 @@ class DoBBA(_BaseClass):
 
     # #### private methods ####
     def _do_bba(self):
+        self.data["bpms2dobba"] = self.bpms2dobba
         tini = _datetime.datetime.fromtimestamp(_time.time())
         print(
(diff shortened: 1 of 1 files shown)


xresende (approved):

fernandohds564 (approved):

apsuite/commisslib/measure_bba.py
```diff
@@ -1562,7 +1562,8 @@ def make_figure_compare_bbas(
 
     # #### private methods ####
     def _do_bba(self):
-        self.data["bpms2dobba"] = self.bpms2dobba
+        if not self.data["bpms2dobba"]:
```
fernandohds564, line 1565: I think this is prone to usage errors, isn't it. I think I understand what you wanted here, but maybe this is not the best way to do it. maybe you could read the measured_bpms list and append the bpms2dobba, what do you think?

PR #308, opened 2025-09-26
FIX AC ORM Disconnection Problem

fernandohds564:
Also, add option for user to choose which LLRF system will be used for RF line measurements in case of conditioning mode is selected.


apsuite/commisslib/meas_ac_orm.py
```diff
@@ -144,15 +181,16 @@ def find_primes(n_primes, start=3):
 class MeasACORM(_ThreadBaseClass):
     """."""
 
-    RF_CONDITIONING_FREQ = 1/12/4e6  # in units of RF frequency
+    RF_CONDITIONING_FREQ = 1 / 12 / 4e6  # in units of RF frequency
     RF_CONDITIONING_VOLTAGEMIN = 60  # [mV]
```
fernandohds564, line 185:
```suggestion
    RF_CONDITIONING_VOLTAGEMIN = 30  # [mV]
```

apsuite/commisslib/meas_ac_orm.py
```diff
@@ -119,13 +146,23 @@ def rf_mode(self, value):
...
+    @property
+    def rf_llrf2use(self):
+        """."""
+        return self._rf_llrf2use
+
+    @rf_llrf2use.setter
+    def rf_llrf2use(self, value):
+        if isinstance(value, str) and value.upper() in {'A', 'B'}:
+            self._rf_llrf2use = value.upper()
+
```
xresende, line 158: I think we should raise an exception here! Otherwise the user may hava a false sense of everything is set correctly.

PR #328, opened 2026-08-18
Mnt fix bug si calc bumps

Gabrielrezende-asc:


apsuite/commisslib/bumps_scans.py
```diff
@@ -0,0 +1,426 @@
...
+        self._initial_state = {}
+        self._initial_state['refx'] = refx
+        self._initial_state['refy'] = refy
+        self._initial_state['bpmxenbl'] = self._bpmxenbl.copy()
+        self._initial_state['bpmyenbl'] = self._bpmyenbl.copy()
+        self._initial_state['fofb_bpmxenbl'] = self._fofb_bpmxenbl.copy()
+        self._initial_state['fofb_bpmyenbl'] = self._fofb_bpmyenbl.copy()
+        self.data['initial_state'] = self._initial_state
+
+    def _is_beam_alive(self):
```
fernandohds564: this name is not the most suitable choice for this method. I wouldn't expect a method with this name to make changes in the machine. Even though it is a hidden method, I think it is better to change this name.

fernandohds564 (approved):

